PR #1, opened 2015-10-15
Fix/initial beats

ph:
Initial version of the logstash-input-beats


lib/logstash/compatibility_layer_api_v1.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.included(base)
+    base.send(:include, InstanceMethods) if self.is_v1?
+  end
+
+  def self.is_v1?
+    LOGSTASH_CORE_VERSION < V2_VERSION
+  end
+  
+  # This allo this plugin to work both in V1 and v2 of logstash-core
```
urso: s/allo/allow

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,215 @@
+# encoding: utf-8
+require "logstash/inputs/base"
+require "logstash/namespace"
+require "logstash/compatibility_layer_api_v1"
+require "lumberjack"
+
+# use Logstash provided json decoder
+Lumberjack::json = LogStash::Json
+
+# Allow Logstash to receive events from Filebeat
```
urso: beats (we've got topbeat, packetbeat and filebeat)

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,215 @@
...
+
+  def ssl_configured?
+    !(@ssl_certificate.nil? || @ssl_key.nil?)
+  end
+
+  def run(output_queue)
+    start_buffer_broker(output_queue)
+
+    while !stop? do
+      # Wrappingu the accept call into a CircuitBreaker
```
urso: s/Wrappingu/Wrapping

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,215 @@
...
+      if @circuit_breaker.closed?
+        connection = @lumberjack.accept # call that creates a new connection
+        next if connection.nil? # if the connection is nil the connection was close.
+        connection = ConnectionDecorator.new(EventDecoration.new,
+                                             @codec.clone, connection)
+        invoke(connection) do |event|
+   	      if stop?
```
urso: formatting gone wrong?

logstash-input-beats.gemspec
```diff
@@ -0,0 +1,33 @@
...
+  s.test_files = s.files.grep(%r{^(test|spec|features)/})
+
+  # Special flag to let us know this is actually a logstash plugin
+  s.metadata = { "logstash_plugin" => "true", "logstash_group" => "input" }
+
+  # Gem dependencies
+  s.add_runtime_dependency "logstash-core", ">= 1.5.4", "< 3.0.0"
+
+  s.add_runtime_dependency "logstash-codec-plain"
+  s.add_runtime_dependency "jls-lumberjack", [">=0.0.24"]
```
urso: bump version. default with latest logstash is 0.0.25

lib/logstash/inputs/beats.rb
ph: +1

lib/logstash/inputs/beats.rb
ph: +1 mergetool probably.

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,215 @@
...
+        end
+      end
+    end
+
+    private
+    def json_event(map, &block)
+      if map.is_a?(Array)
+        map.each { |item| json_event(item, &block) }
+      else
+        line = map.delete("line")
```
urso: can we make this one configurable? Right now on beats side all fields named 'message' are renamed to 'line' to run codec.

Is it ok to have "line" still in the final event. So filebeat can publish 'line' instead of 'message' for to elasticsearch. The 'message' field will be added by codec and named either 'line' or 'message'. should be user configurable.

ph: I though it was the other way arround? That beats were using "message" as the main field and was renaming it to line before sending it to logstash?

urso: Yes, that's right. in beats the 'message' field is renamed to "line" right now. Once the codec is configurable I would like to use 'line' in filebeat again as default line output (like LSF does). So elasticsearch will see 'line' and not 'message'.

With filebeat potentially writing to elasticsearch instead of logstash, file encoding might be dealt with in filebeat.

ph:
I was able to install the plugin in:
Logstash 1.5.4 and Logstash 2.0

ph:
require https://github.com/elastic/ruby-lumberjack/pull/22

ph:
@urso Can you give this a shot?

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,173 @@
...
+
+  public
+  def stop
+    @lumberjack.close
+  end
+
+  private
+  def create_event(codec, map)
+    # Filebeats uses the `message` key and LSF `line`
+    message = map.delete("message")
```
urso: what happens if map does not contain message?

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,173 @@
...
+
+    @codec.decode(message) do |decoded|
+      decorate(decoded)
+      map.each { |k, v| decoded[k] = v; v.force_encoding(Encoding::UTF_8) }
```
urso: v might not be a string. can be a bool, a number, an array or another object. Does the json decoder ensure we're dealing with utf8 always?

lib/logstash/inputs/beats.rb
```diff
@@ -0,0 +1,173 @@
...
+    end
+  end
+
+  private
+  def invoke(connection, codec, &block)
+    @threadpool.post do
+      begin
+        # If any errors occur in from the events the connection should be closed in the
+        # library ensure block and the exception will be handled here
+        connection.run do
```
urso: connection.run do |map|
        ...
    end

lib/logstash/inputs/beats.rb
ph: I think this is a good assumption to make that the json decoder will keep the correct encoding.

lib/logstash/inputs/beats.rb
ph: fixed.

PR #10, opened 2015-10-28
parse and set @timestamp if in original message

urso:
we plan to always send '@timestamp'. So filters have to update '@timestamp' if required. Timestamps set by codec will be ignored by this change

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -156,7 +156,6 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   rescue LogStash::TimestampParserError => e
     LOGGER.warn("Error parsing @timestamp string, setting current time to @timestamp",
       :value => ts.inspect, :exception => e.message)
-    end
   end
 
   private
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require "logstash/inputs/base"
 require "logstash/namespace"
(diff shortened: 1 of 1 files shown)


lib/logstash/inputs/beats.rb
```diff
@@ -142,6 +146,20 @@ def create_event(codec, map)
...
+    return nil if ts.nil?
+    timestamp = LogStash::Timestamp.coerce(ts)
+    return timestamp if timestamp
+
+    LOGGER.warn("Unrecognized @timestamp value, setting current time to @timestamp",
+      :value => ts.inspect)
+  rescue LogStash::TimestampParserError => e
+    LOGGER.warn("Error parsing @timestamp string, setting current time to @timestamp",
+      :value => ts.inspect, :exception => e.message)
+    end
```
ph: When I remove this `end` I can run the test, if not i get a syntax error.

ph:
Small typo and I am okay to merge this.

urso:
still needs some thought. Is the implementation okay for the default behavior? Some codecs e.g. json might set @timestamp if this field is present in message, but we would overwrite the field with our own value. On the other hand, why would someone want to use the 'json' codec if similar (better) behavior is supported by json filter?

ph:
@urso I think this is okay for a first implementation, the json give it a bit of flexibility because this filter doesn't expect to fully replace the whole document but instead take a field as a source of "stringified json" and extract the content in a user supplied field.

And since this is a plugin we can release a new version easily if this behavior doesn't work as intended.

urso:
+1

PR #15, opened 2015-11-04
Make sure the `add_field` and the `tags` options from the config are popagated

### LINKED ISSUE #12: "add_field" and "tags" configuration options have no effect
I am using this configuration with logstash 2.0.0 ...
```
input {
  beats {
    port => 5044
    add_field => { "application" => "beats" }
    tags => [ "beats" ]
  }
}

output {
  stdout {
    codec  => rubydebug {
      metadata => true
    }
  }
}
```
... and send some data using 'topbeat'

In the output I see neither the field 'application' nor the tag 'beats'. This would be helpful to create filters for this input when other inputs are used in parallel

ph:
When an event was sent to logstash from topbeats logstash wasn't decorating the event to add the `add_field` and the `tags` from the configuration.

Fix #12

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -137,10 +137,10 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
     else
       # All codecs expects to work on string
       @codec.decode(target_field.to_s) do |decoded|
-        decorate(decoded)
         ts = coerce_ts(map.delete("@timestamp"))
         decoded["@timestamp"] = ts unless ts.nil?
         map.each { |k, v| decoded[k] = v }
+        decorate(decoded)
         return decoded
       end
     end
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -125,13 +125,15 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
     @lumberjack.close
   end
 
-  private
+  public
   def create_event(codec, map)
     # Filebeats uses the `message` key and LSF `line`
     target_field = target_field_for_codec ? map.delete(target_field_for_codec) : nil
 
     if target_field.nil?
-      return LogStash::Event.new(map) 
+      event = LogStash::Event.new(map) 
+      decorate(event)
+      return event
     else
       # All codecs expects to work on string
       @codec.decode(target_field.to_s) do |decoded|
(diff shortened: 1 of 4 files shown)


ph:
@urso might want to take a look?

I also changed the LOGGER to @logger also created https://github.com/logstash-plugins/logstash-input-beats/issues/16

lib/logstash/inputs/beats.rb
```diff
@@ -125,16 +125,17 @@ def stop
     @lumberjack.close
   end
 
-  private
+  public
```
urso: why public? for testing?

urso:
What happens if event already has field 'tags' ? Will these be overwritten.

urso:
why change LOGGER to @logger ? This backwards compatible to 1.5.4 ?

ph:
We never user a `LOGGER` constant in any plugins, I not sure it even worked before.

Tags should be merged, but the code base expect an array, so this should be fixed in core. https://github.com/elastic/logstash/blob/master/lib/logstash/util/decorators.rb#L39

lib/logstash/inputs/beats.rb
ph: Oh my, you saw this, all theses blocks was making my head explode ;)

ph:
Also added a check for using `add_field` with a metadata and it work.

urso:
Code seems to behave different if target_field is available or not available.

If target_field is available:
1. event is generated from all input hash
2. then fields are merged into event
3. then tags are merged into event["tags"] (if events["tags"] is array, otherwise exception is thrown)

If target_field is not available
1. event with '@timestamp', '@version' and 'message' fields is generated
2. configured tags and fields are added to event
3. original fields are added to event (potentially overwriting plugin configured fields and tags, no merging)

For constistent behavior decorate should be [moved after this line](https://github.com/logstash-plugins/logstash-input-beats/pull/15/files#diff-28d8b79a78399c1304c58e81cd3672f2R143).

Still open (wouldn't say it's critical): what to do if 'tags' is given in event but no array (e.g. comma separated string)?

ph:
@urso correct good catch I'll make sure both scenario behave the same.
For tags and things, I'll create another issue for core since all plugins could have the same problem.

ph:
@urso I have changed where the event decoration takes places. test still passes.

Concerning comma separated tags, I think the safest way would be to take the whole string and generate an array with only one element?

```
Tags => "foo,bar,foobar"
```
Would become this
```
Tags => ["foo,bar,foorbar"]
```

Any scenario in mind that could generate tags without an array?

urso:
decoration LGTM

for tags the beats also send an array, so should be no issue. In worst case one can still use fields and logstash filters to massage some events.

PR #22, opened 2015-11-12
Add support for `identity_stream`

### LINKED ISSUE #13: Multiline codec can't differentiate different sources
Currently the Beats input doesn't natively support multiline events from multiple different streams. As multiline codec is the thread safe option, this should be fixed accordingly.

Same issue as in the file input: logstash-plugins/logstash-input-file#44

ph:
This PR add support for using an `identity_stream`, the stream is useful if you want to use a codec that buffer data before creating new events like the `Multiline Codec`  and it solve the problem of merging different sources.

This code use multiples keys transmitted from beats to create a unique stream.
In the case of a filebeat frame we will use this structure if possible, we are ignoring `nil` keys when generating the identity. 

`host, ip, beats.name,type, input_type, source`

Fixes #13


ph:
@suyograo Since the identity class is only available in version 2.0 of the multiline codec, the input beats for 1.5 will not support this feature are we OK with that?

suyograo:
@ph yes that should be ok

urso:
@suyograo @ph @tbragin not sure I'm ok with breaking BC. For beats we said we wanted to be BC to 1.5 at least. As this breaks BC we have to guarantee a way critical fixes/changes should be available for both version, right? Otherwise we have to force all beats users (e.g. migrating from LSF) to upgrade logstash too.

ruflin:
As far as I understand, that is a feature / issue that also didn't exist for lsf. My main concern is if we will have two versions of logstash-input-beats in the future? So if we have a bug that is fixed and a user on 1.5 updates the plugin, will he also get a new version?

ph:
@guyboertje added a few missing commits.

ph:
@urso What do you think about using theses fields for creating a unique identity, I think this is mostly needed with filebeat.

https://github.com/logstash-plugins/logstash-input-beats/pull/22/files#diff-b8da80c241831da57d7b1c09674a3160R399

lib/lumberjack/beats/server.rb
```diff
@@ -375,17 +380,29 @@ def close
...
       @fd.syswrite(@ack_handler.ack_frame(sequence))
     end
+
+    def version_1?
+      @version == Parser::PROTOCOL_VERSION_1
+    end
+
+    def identity_stream(map)
+      [peer,
+       map.fetch("beats", {})["name"],
```
ruflin: @ph Do you assume name equals to hostname or the manual set name? We renamed it recently to beat.hostname but some discussions are going on to change it again.

ph: Well, Actually it can be the default or the user configured value, it's just more 'data' to identity a stream.  The source and the first part of the stream with `peer` should be pretty unique since it contains the IP + connected port. 

is that too much/enough?

WDYT?

lib/lumberjack/beats/server.rb
```diff
@@ -375,17 +380,29 @@ def close
...
+       map["type"],
+       map["input_type"],
+       map["source"]].compact.join("-")
```
urso: not sure about these fields for stream identity.
See fields documentation for [packetbeat](https://github.com/elastic/packetbeat/blob/master/etc/fields.yml), [filebeat](https://github.com/elastic/filebeat/blob/master/etc/fields.yml) and [topbeat](https://github.com/elastic/filebeat/blob/master/etc/fields.yml) .

Only common fields you will get from all beats:
  @metadata.beats,
  @metadata.type   (which is similar to map["type"]
  beat.name

These fields are not enough to build a unique ID for filebeat based logs though.
filebeat specific:
   input_type  (not sure if really required)
   source

Not sure if stream identifier can be of interest for packetbeat based data. Topbeat likely not, as topbeat is about some metrics.

[winlogbeat](github.com/elastic/winlogbeat) might export other [fields](https://github.com/elastic/winlogbeat/blob/master/eventlog/eventlog.go#L92). No idea if multiline will be of interest to windows event logs. Due to winlogbeat potentially collecting logs from remote hosts, you must also include:
  - computerName
  - sourceName

@andrewkroh might have some more input on unique stream identifier for winlogbeat.

guyboertje:
@ph - i am checking this locally.  I am worried about the eviction flush. We may need an alternative mechanism to flush an event to the pipeline.

lib/lumberjack/beats/server.rb
ph: @urso but the current logic for creating the identity stream with removed any field that return nil, so the length of it will depends on the shipper?

ph:
I have sync with @ruflin and we are looking at generating the `identity_stream` information on the filebeat side instead of doing in the input. This would make sense because the source of truth is on the other side.

ruflin:
Here my thoughts on this after the sync with @ph:

Logstash needs to uniquely identify the filebeat instance and the file and event was sent from to correlate log lines. Hostname + Filepath is not unique because of two reasons:
* Multiple filebeat instances can run on one host
* Path can change over time (log rotation).

Log line correlation is filebeat specific and not a general beats problem. Based on this the following suggestion for a solution:

* Unique beat identifier independent of hostname
* Unique file identifiers as md5 (or similar)

The unique identifier for each beat would have to be implemented in libbeat. It could be a UUID in `beat.id`. This could also be used later for other features where a beat must be identified uniquely. The unique md5 value of a file would be generated based on the unique file identifiers by filebeat (2 under Linux, 3 under Windows) and sent as part of the event as `fileId`.

The changes on filebeat and libbeat side would probably be made after GA. On the logstash side I would suggest something like:

```
if exist(beat.id) && (fileId) {
   streamId = beat.id + fileId
} else {
   streamId = beat.name + source
}
```

For non filebeat beats, streamId should not be needed.

guyboertje:
@ph and I discussed an alternative API to allow arbitrary flushing of codec buffered lines into an event.
We will add this as a new Issue.

We also discussed flushing the codec when a connection is closed.

@ph - regarding flushing when the connection closes, you can do this for now...
```ruby
@codec.codec_without_usage_update(identity).flush do |event|
  decorate_and_enqueue(event)
end
```

guyboertje:
If NOT implementing connection close - LGTM

ph:
@ruflin +1 on this, this will make it seamless to work in future versions.

ph:
@guyboertje 
I don't think this code will work, because when a disconnect happen the notion of identity is gone because it's tied to an event.

guyboertje:
@ph - really? So how will we know that a 'stream' is done with, finito, dead (not just pining for the fjords)?
Will the beat tell us? ATM, if the 'stream' has not sent new data, after one hour it is considered inactive and the eviction process will try to flush the buffered data and fail.

ph:
The way the multiline input work, it need new data to correctly do his work. If there is is no new data coming in the codec data will be stuck in the codec. This scenario will mostly happen at the end of files and may occur data loss.

This behavior isn't new with this added identity stream.

But we have more rejoicing, The identityMapCodec has an eviction mechanism to clear out unused stream, but to correctly work it require a proc that can push to the queue and that will be called when the eviction is done.

The current code doesn't allow the proc to be safely call outside of the normal flow and it need some refactor.

I will create that change in another PR so this feature isn't blocked.

urso:
@ruflin @ph I see streamID possibly of interest to winlogbeat ... maybe other beats in future (syslogbeat?). How about using **resourceID** instead of **fileID** ?

urso:
We will not generate additional events like 'new resource', 'resource gone' in beats. I think this is a general problem for logstash. I don't have a better solution other then timeouts. But timeouts can be quite small for multiline-cache, as multi-line events should be relatively close in time.

urso:
All in all LGTM

ph:
@urso resource_id (with the underscore) make more sense and less tied to the actual beats and more to how we handle the event. @ruflin Are you OK with that?

Concerning *new resource/resource gone* events, I think we should not do that and keep the communication as stateless as possible.

urso:
+1

PR #27, opened 2015-11-18
Correctly merge multiline events with the codec

ph:
The problem was the buffer wasn't cleared correctly when received a new
event header resulting in duplicates of lines. Added a test to cover it

Fixes #22

### DIFF
--- a/spec/inputs/beats_spec.rb
+++ b/spec/inputs/beats_spec.rb
@@ -91,10 +91,11 @@ describe LogStash::Inputs::Beats do
 
       context "without a `target_field` defined" do
         it "decorates the event" do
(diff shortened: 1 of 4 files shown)


spec/inputs/beats_spec.rb
```diff
@@ -113,8 +115,41 @@
         let(:event_map) { {"message" => "hello?", "tags" => ["syslog"]} }
 
         it "retuns nil" do
```
guyboertje: s/retuns/returns/

guyboertje:
LGTM

jordansissel:
LGTM

PR #28, opened 2015-11-18
Copies the beat.hostname field into host

### LINKED ISSUE #25: Copy beat.hostname into host
The Beats are sending the hostname on which they run in the `beat.hostname` field. In Logstash, most input plugins are using the `host` field for this purpose. This is annoying for people that use Logstash with a lot of different sources.

After a bit of discussion on the topic, we decided that the best way forward would be to keep using `beat.hostname` in the Beats but have the logstash-input-beats plugin copy it to the `host` field.

tsg:
Does this before the codec is applied, so if a "host" field is the result of parsing, that one will override it. Closes #25.

@ph Can you have a look at this one, please? I don't really know what I'm doing in ruby :-).

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -128,6 +128,12 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   public
   def create_event(map, identity_stream)
+    # Copy beat.hostname into host
+    host = map.fetch("beat", {})["hostname"]
+    if host
+      map["host"] = host
+    end
+
     # Filebeats uses the `message` key and LSF `line`
     target_field = target_field_for_codec ? map.delete(target_field_for_codec) : nil
 
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -128,17 +128,13 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   public
   def create_event(map, identity_stream)
-    # Copy beat.hostname into host
-    host = map.fetch("beat", {})["hostname"]
-    if host
-      map["host"] = host
-    end
 
     # Filebeats uses the `message` key and LSF `line`
     target_field = target_field_for_codec ? map.delete(target_field_for_codec) : nil
 
     if target_field.nil?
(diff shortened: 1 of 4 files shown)


tsg:
@ph I added another commit. I changed the tactics a bit, in doing my copying after the codec is applied. This is because I previously assumed that the JSON codec would overwrite existing fields, but that wasn't the case in this particular implementation.

lib/logstash/inputs/beats.rb
```diff
@@ -148,6 +151,16 @@ def create_event(map, identity_stream)
     return nil
   end
 
+  # Copies the beat.hostname field into the host field unless
+  # the host field is already defined
+  private
+  def copy_beat_hostname(event)
+    host = event["beat"] ? event["beat"]["hostname"] : nil
+    if host and not event["host"]
```
ph: We don't use the and/not statements ;)

```ruby
if host && event["host"].nil?
```

ph:
jenkins test this please

ph:
jenkins, test it

spec/inputs/beats_spec.rb
```diff
@@ -117,6 +118,44 @@
...
+      end
+
+      context "with a host field in the message" do
+        let(:codec) { LogStash::Codecs::JSON.new }
+        let(:event_map) { {"message" => '{"host": "linux02"}', "beat" => {"hostname" => "linux01"} } }
+
+        it "should take the host from the JSON message" do
+          event = beats.create_event(event_map, identity_stream)
+          expect(event["host"]).to eq("linux02")
+        end
```
ph: There is only small problem with the tests. We use the `message` field to check if we should apply a codec or not to the event and theses tests will only trigger the `else` part of this conditionals. https://github.com/logstash-plugins/logstash-input-beats/pull/28/files#diff-28d8b79a78399c1304c58e81cd3672f2R140

tsg: Oh, good point, I added a test that doesn't have the `message` field.

tsg:
Comments should be addressed now. Let me know if you want me to rebase or squash.

urso:
This PR will be required to work with Logstash 1.5 and Logstash 2.0 branches, right?

Issue #37, opened 2015-12-29
Threading issue under high volume, ThreadError: Mutex is not locked


ph:
```
:message=>"Beats input: unhandled exception", :exception=>#<ThreadError: Mutex is not locked>, :backtrace=>["org/jruby/ext/thread/Mutex.java:106:in unlock'", "org/jruby/ext/thread/Mutex.java:151:insynchronize'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-lumberjack-2.0.5/lib/logstash/sized_queue_timeout.rb:27:in `push'"...
```

This also raise this error `#<ConcurrencyError: No message available>,` which in turn do a  Concurrent::RejectedExecutionError.

A few notes the sizequeue from the logstash input lumberjack seems to conflict with the one in the beats input. This shouldn't be a big problem since both code is the same but it may cause issue in debugging or updating, I will rename it the separation clearer.

Concerning the bug, I believe there is a an issue with the `ConditionVariable` inside the SizeQueueWith timeout, when a lot of servers are connected to logstash and the platform is under high load.

I think the best solution is to remove the custom code here and rely on Java Queue which already support a timeout values or by removing the timeout completely. This will have the benefit of reducing the custom code and make sure the class is free of any concurrent bug.

I haven't been able to reproduce it locally, but I still think removing that code an use the one shipped with jruby is the best bet here. 

A small story, I was thinking about using it before, but before we were avoiding using pure java classes.

PR #38, opened 2016-01-04
Refactoring of the plugin handling of connection

ph:
This PR introduces a few notables changes to the way the plugin is
structured.

- It uses the Callback approch proposed by @guyboertje instead of using
  a block.

- It decouples the connection handing and the data parsing into
  separates class easiy unit testable.

- It uses the Java Synchronous queues instead of my homebrew version
which had a concurrency bug under high load. (Note logstash core will
also use the same queue)

- It fix an issue with incorrectly parsed date when applying the payload data to
  a decoded message frame.

- It had better flush handling when we shutdown logstash, codec content
  will be send to the queue.

- it keep track of the connections and remove them when problems
  happens.

- When a connection is close from the library the input plugin will be
  aware of it and be able to flush the content of the codec.

**Reviewer**
- Missing an integration test for the multiline case, working on that.
- @guyboertje Can you take a look at the codec/flush/callback code?
- Real world test would be appreciated :heart: 

Fixes: #37


ph:
@pickypg This should fix the `concurrency` error we saw, this PR also introduce code quality improvement,more tests and life quality improvements.

ph:
Another thing to note I have added more logging to the input when it run in debug mode,
Also the thread name will now match the connected host to help us debug future problems.

jsvd:
got an exception running the tests, but haven't been able to reproduce with the same seed:

```
% bundle exec rspec
Using Accessor#strict_set for specs
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
.........F..............................................................................

Failures:

  1) A client using plain tcp connection should fail to communicate to ssl based server
     Failure/Error: expect {
       expected RuntimeError, got #<Errno::ECONNRESET: Connection reset by peer - Connection reset by peer> with backtrace:
         # ./lib/lumberjack/beats/client.rb:157:in `read_version_and_type'
         # ./lib/lumberjack/beats/client.rb:145:in `ack'
         # ./lib/lumberjack/beats/client.rb:134:in `write_sync'
         # ./lib/lumberjack/beats/client.rb:42:in `write'
         # ./spec/integration_spec.rb:137:in `(root)'
         # ./spec/integration_spec.rb:132:in `(root)'
         # /Users/joaoduarte/.rvm/gems/jruby-1.7.22/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'
     # ./spec/integration_spec.rb:132:in `(root)'
     # /Users/joaoduarte/.rvm/gems/jruby-1.7.22/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'

Finished in 31.2 seconds (files took 6.81 seconds to load)
88 examples, 1 failure

Failed examples:

rspec ./spec/integration_spec.rb:131 # A client using plain tcp connection should fail to communicate to ssl based server

Randomized with seed 6003

% SPEC_OPTS="-s 6003" bundle exec rspec
Using Accessor#strict_set for specs
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
........................................................................................

Finished in 29.99 seconds (files took 7.13 seconds to load)
88 examples, 0 failures

Randomized with seed 6003
```

Likely a timing issue when closing a socket?

lib/logstash/inputs/beats.rb
```diff
@@ -13,7 +17,25 @@
 #
 # https://github.com/elastic/filebeat[filebeat]
 #
+
+class LogStash::Codecs::Base
```
jsvd: maybe explain here why this is needed?

jsvd:
Got an Thread death but didn't influence the output:

```
% SPEC_OPTS="-s 6003" bundle exec rspec
Using Accessor#strict_set for specs
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
.........................................
Exception: java.lang.ThreadDeath thrown from the UncaughtExceptionHandler in thread "Thread-85"
...............................................

Finished in 31.24 seconds (files took 6.58 seconds to load)
88 examples, 0 failures

Randomized with seed 6003
```

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
     @threadpool.post do
-      while !stop?
-        @output_queue << @buffered_queue.pop_no_timeout
+      LogStash::Util.set_thread_name("[beats-input]-buffered-queue-broker")
+
+      begin
+        while !stop?
+          @output_queue << @buffered_queue.take
+        end
+      rescue => e
```
jsvd: if we know the exception, maybe we should do
```ruby
...
rescue InterruptionException
  @logger.error("Beats input: bufferered queue broker exception", :exception => e) unless stop?
rescue => e
  @logger.error("Beats input: unexpected exception, :exception => e)
end
```

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
     end
   end
+
+  # Handle the data coming from a connection
+  # Decide which Process should be used to decode the data coming
+  # from the beat library.
+  #
+  # - Should we use a codec on specific field?
+  # - Should we just take the raw content of the parsed json frame
+  class ConnectionHandler
```
jsvd: may I suggest placing this in another file? same with the transforms?

jsvd:
I may have done something wrong. but couldn't get data from filebeat:

```
% bin/logstash -f beats.config
Settings: 
Logstash startup completed
Beats input: unhandled exception {:exception=>#<NoMethodError: undefined method `accept' for #<LogStash::Codecs::IdentityMapCodec:0x2da696ac>>, :backtrace=>["/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:278:in `process'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:253:in `accept'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:420:in `data'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:399:in `read_socket'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:411:in `ack_if_needed'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:395:in `read_socket'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:255:in `json_data_payload'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:172:in `feed'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:305:in `compressed_payload'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:172:in `feed'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:380:in `read_socket'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/lumberjack/beats/server.rb:355:in `run'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:253:in `accept'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:177:in `handle_new_connection'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-lumberjack-2.0.5/lib/logstash/circuit_breaker.rb:42:in `execute'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:177:in `handle_new_connection'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:130:in `run'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/concurrent-ruby-0.9.2-java/lib/concurrent/executor/executor_service.rb:515:in `run'", "Concurrent$$JavaExecutorService$$Job_290572464.gen:13:in `run'"], :level=>:error}
```

logstash config:

```
input {
  beats { port => 5044 }
}
output {
  stdout { codec => rubydebug }
}
```

filebeat.yml:

```yaml
filebeat:
  # List of prospectors to fetch data.
  prospectors:
    -
      paths:
        - /tmp/log
      input_type: log
      document_type: log
      scan_frequency: 1s

output:
  logstash:
    hosts: ["localhost:5044"]
    worker: 1
    loadbalance: true
  console:
    pretty: true

shipper:

logging:
  files:
    rotateeverybytes: 10485760 # = 10MB
```

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
+      
+      @json_transformer = JsonTransform.new(@input)
+      @codec_transformer = CodecTransform.new(@input)
+    end
+
+    def accept
+      @logger.debug("Beats input: waiting from new events from remote host",
+                    :peer => @connection.peer)
+
+      @connection.run { |map, identity_stream| process(map, identity_stream) }
```
jsvd: I understand that in other languages and paradigms a hashmap is often shortened to map, but in ruby it's more common to call it a hash.
What do you think about consistently naming this hash instead of map?

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
+  end
+
+  # Take the the raw output from the library, decorate it with
+  # the configured tags in the plugins and normalize the hostname
+  # for backward compatibility
+  #
+  #
+  # @see [Lumberjack::Beats::Parser]
+  #
+  class JsonTransform < EventTransformCommon
```
jsvd: So..JsonTransform doesn't transform Json?

guyboertje:
@ph @jsvd - the gemspec dep for multiline codec should be `~> 2.0.5`

lib/logstash/inputs/beats.rb
guyboertje: Maybe `NoCodecTransform` would be a more sensible name?

jsvd:
Updating the multiline codec made it work, thanks @guyboertje 

@ph is it normal for the hostname to appear 3 times in the event?

```
{
       "message" => "hey",
      "@version" => "1",
    "@timestamp" => "2016-01-06T11:55:48.062Z",
          "beat" => {
        "hostname" => "Joaos-MBP.lan",
            "name" => "Joaos-MBP.lan"
    },
         "count" => 1,
        "fields" => nil,
    "input_type" => "log",
        "offset" => 24,
        "source" => "/tmp/log",
          "type" => "log",
          "host" => "Joaos-MBP.lan",
          "tags" => [
        [0] "beats_input_codec_applied"
    ]
}
```

jsvd:
Doing a ctrl+c while receiving events got me this:

```
^CSIGINT received. Terminating immediately.. {:level=>:fatal}
Logstash shutdown completed
Beats input: unhandled exception {:exception=>#<ConcurrencyError: interrupted waiting for mutex: null>, :backtrace=>["org/jruby/ext/thread/Mutex.java:94:in `lock'", "org/jruby/ext/thread/Mutex.java:147:in `synchronize'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-lumberjack-2.0.5/lib/logstash/circuit_breaker.rb:89:in `state'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-lumberjack-2.0.5/lib/logstash/circuit_breaker.rb:47:in `execute'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:177:in `handle_new_connection'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:130:in `run'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/concurrent-ruby-0.9.2-java/lib/concurrent/executor/executor_service.rb:515:in `run'", "Concurrent$$JavaExecutorService$$Job_290572464.gen:13:in `run'"], :level=>:error}
```

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
+  class EventTransformCommon
+    def initialize(input)
+      @input = input
+      @logger = input.logger
+    end
+
+    # Copies the beat.hostname field into the host field unless
+    # the host field is already defined
+    def copy_beat_hostname(event)
+      host = event["beat"] ? event["beat"]["hostname"] : nil
```
jsvd: specially with the new java event, `event["beat"]["hostname"]` should become an anti pattern, to be refactored to `event["[beat][hostname]"]`

lib/lumberjack/beats/server.rb
```diff
@@ -66,8 +66,16 @@ def run(&block)
         # unless the server is closing
         next unless connection
 
+        Thread.abort_on_exception = true
         Thread.new(connection) do |connection|
-          connection.run(&block)
+          begin
+            connection.run(&block)
+          rescue Lumberjack::Beats::Connection::ConnectionClosed
```
jsvd: in what situation does this happen? if a client terminates the connection?

lib/lumberjack/beats/server.rb
```diff
@@ -66,8 +66,16 @@ def run(&block)
...
         next unless connection
 
+        Thread.abort_on_exception = true
         Thread.new(connection) do |connection|
-          connection.run(&block)
+          begin
+            connection.run(&block)
+          rescue Lumberjack::Beats::Connection::ConnectionClosed 
+            # Connection will raise a wrapped exception upstream, 
+            # but if the threads are managed by the library we can simply ignore it.
```
jsvd: I'm not very familiar with the insides so the "if the threads are managed...we can simply ignore it" is strange, what if they're not? also, by library you mean ruby-lumberjack?

logstash-input-beats.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency "logstash-codec-plain"
   s.add_runtime_dependency "concurrent-ruby", "~> 0.9.2"
+  s.add_runtime_dependency "thread_safe"
   s.add_runtime_dependency "logstash-codec-multiline", "~> 2.0.3"
```
jsvd: bump to "~> 2.0.5"

lib/lumberjack/beats/server.rb
```diff
@@ -322,9 +358,11 @@ def run(&block)
       OpenSSL::SSL::SSLError,
       IOError,
       Errno::ECONNRESET,
-      Errno::EPIPE
+      Errno::EPIPE,
```
jsvd: I suggest putting these exceptions on a CONSTANT an just rescue *CONSTANT => e

lib/lumberjack/beats/server.rb
```diff
@@ -66,8 +66,16 @@ def run(&block)
         # unless the server is closing
         next unless connection
 
+        Thread.abort_on_exception = true
```
jsvd: this is a bit scary setting it here.
by default logstash enables abort on exception so this shouldn't be needed.

if it's really necessary I suggest adding an extra check that, if it's false, log a warning that it's being enabled

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
+      @input = input
+      @queue = queue
+      @logger = input.logger
+
+      # We need to clone the codec per connection, so we can flush a specific 
+      # codec when a connection is closed.
+      @codec = input.codec.dup
+      
+      @json_transformer = JsonTransform.new(@input)
+      @codec_transformer = CodecTransform.new(@input)
```
jsvd: Not a big fan of this name :(

from what I understand, these are 2 transformers for the scenarios where the codec was and wasn't applied, right?

lib/logstash/inputs/beats.rb
```diff
@@ -207,9 +210,229 @@ def invoke(connection, &block)
...
+  # from beats, apply the configured tags, normalize the host and try to coerce
+  # the timestamp if it was provided in the map.
+  class CodecTransform < EventTransformCommon
+    def transform(event, map)
+      ts = coerce_ts(map.delete("@timestamp"))
+
+      event["@timestamp"] = ts unless ts.nil?
+      map.each { |k, v| event[k] = v }
+      super(event)
+      event.tag("beats_input_codec_applied")
```
jsvd: in the event of a success is it necessary to apply the tag? maybe follow the grok/date convention and only apply on failure?
Also, "beats_input_codec_applied" seems there's a beats input codec, which isn't really correct, more like "input_codec_sucess". Or somehow fetch the name of the plugin class in order to have "input_json_codec_ok"?

lib/lumberjack/beats/server.rb
ph: The ruby-lumberjack has 2 way to work, which is a bit unfriendly:

-  it can manage the thread for you but does it in a really minimal way
- or you can do all the managing yourself (which is what the input is doing) 

But for integration testing for the library its handy to have a way to let it do all the heavy lifting

lib/lumberjack/beats/server.rb
ph: good catch this shouldnt be here, I was debugging something :(

lib/logstash/inputs/beats.rb
ph: I think its better in that case to add it all the time, since it will give more visibility into what the plugins did.

I wanted to expose if the beats input went through a json map or a codec decode since its not obvious all the time, but I am + 1 to add the name of the codec used in the tag.

lib/logstash/inputs/beats.rb
ph: > from what I understand, these are 2 transformers for the scenarios where the codec was and wasn't applied, right?

Correct, since the logic is kinda different in both case.

> Not a big fan of this name :(
The variable, the classes or both? Any suggestion ? :(

lib/lumberjack/beats/server.rb
ph: Correct, anything that can terminate the connection will raise this exception and wrap the original exception if further action are needed.

lib/logstash/inputs/beats.rb
ph: I would +1 this but the way our plugins are structured doesn't make it easy, theses classes should be inside a "LogStash::Inputs::Beats" namespace which we can't do because we have a class named beats. I wish the plugin structure wasn't directed by the logstash main structure.

Would you be OK to group theses classes inside a `BeatsSupport` namespace?

lib/logstash/inputs/beats.rb
ph: +1 will make it more targeted.

logstash-input-beats.gemspec
ph: +1

lib/lumberjack/beats/server.rb
ph: +1

lib/logstash/inputs/beats.rb
ph: `NoCodecTransform` +1

lib/logstash/inputs/beats.rb
ph: I don't have a preference, I can rename it hash

ph:
> @ph is it normal for the hostname to appear 3 times in the event?

Its normal,

Beats always send the `hostname` and the `name` of the user configured shipper which fallback to the hostname if nothing is set.

`host` is for logstash backward compatibility, some plugins use it internally and the `Event#to_s` also use it by default.

ph:
> Beats input: unhandled exception {:exception=>#<ConcurrencyError: interrupted waiting for mutex: null>, :backtrace=>["org/jruby/ext/thread/Mutex.java:94:in `lock'", "org/jruby/ext/thread/Mutex.java:147:in `synchronize'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-lumberjack-2.0.5/lib/logstash/circuit_breaker.rb:89:in `state'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-lumberjack-2.0.5/lib/logstash/circuit_breaker.rb:47:in `execute'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:177:in `handle_new_connection'", "/Users/joaoduarte/new_plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:130:in `run'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/concurrent-ruby-0.9.2-java/lib/concurrent/executor/executor_service.rb:515:in `run'", "Concurrent$$JavaExecutorService$$Job_290572464.gen:13:in `run'"], :level=>:error}

I will check this exection, nothing was changed for that part of the code.

ph:
> Likely a timing issue when closing a socket?

Sound like it too :panda_face:

lib/logstash/inputs/beats.rb
ph: +1

lib/lumberjack/beats/server.rb
ph: On the plugin side we need to know when something close a connection, before that we were simply ignoring it and closing the socket. no feedback was ever given back to the library which mean we could have lost events inside a codec. :(

guyboertje:
Other than agreeing with @jsvd comments, this LGTM.

lib/lumberjack/beats/server.rb
```diff
@@ -132,6 +140,7 @@ class Parser
     PROTOCOL_VERSION_2 = "2".ord
 
     SUPPORTED_PROTOCOLS = [PROTOCOL_VERSION_1, PROTOCOL_VERSION_2]
+    class UnsupportedProtocol < Exception; end
```
ph: StandardException

lib/logstash/inputs/beats.rb
ph: Went through everything renamed map -> hash tests passes hurray.

lib/logstash/inputs/beats.rb
ph: +1, I have changed it.

lib/lumberjack/beats/server.rb
jsvd: may I suggest for your first comment about ruby-lumberjack's 2 ways of working to be included somewhere in the code base?

ph: I will add a comment in the `server.rb`, I want to refactor the library later on.

lib/logstash/inputs/beats.rb
jsvd: Both...maybe... "SuccessfulDecodeTransformer" and "FailedDecodeTransformer"?

ph: Well, There isn't a concept of failed or successful in that part of the code?
They are just different branch of a conditionals, different data handling?

jsvd: oops I misread the code..
what about RawEventTransform and DecodedEventTransform?

ph: I am fine with that suggestion. ^

jordansissel:
The specs pass most of the time for me, but I got it to get stuck once in rspec but couldn't get the seed or the stuck test out of it. Could be a timing problem?

ph:
@jordansissel could, do you have the rspec doc output?

ph:
@jsvd the mutex:null isn't new, trying to find where the `InterruptionException` is made, its probably originating from the Socket blocking read call.

jordansissel:
@ph I don't, but I'm trying to reproduce and using doc output to catch the sticking test.

jordansissel:
Got a new failure. Seed 3.

```
Failures:

  1) LogStash::Inputs::Beats#handle_new_connection when an exception occur calls flush on the handler and tag the events
     NoMethodError:
     # ./lib/logstash/inputs/beats.rb:203:in `handle_new_connection'
     # ./spec/inputs/beats_spec.rb:110:in `(root)'
     # /home/jls/.rvm/gems/jruby-1.7.23/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'

Finished in 26.17 seconds (files took 2.59 seconds to load)
88 examples, 1 failure

Failed examples:

rspec ./spec/inputs/beats_spec.rb:107 # LogStash::Inputs::Beats#handle_new_connection when an exception occur calls flush on the handler and tag the events

Randomized with seed 3
```

Rerunning doesn't reproduce it, so it may be a timing issue?

ph:
@jordansissel visibility issue across thread, between the `#run` and the `#handle_connection`

jordansissel:
```
Failures:

  1) LogStash::Inputs::Beats#handle_new_connection when an exception occur calls flush on the handler and tag the events
     NoMethodError:
     # ./lib/logstash/inputs/beats.rb:203:in `handle_new_connection'
     # ./spec/inputs/beats_spec.rb:110:in `(root)'
     # /home/jls/.rvm/gems/jruby-1.7.23/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'

  2) LogStash::Inputs::Beats#register raise no exception
     Failure/Error: plugin.run(pipeline_queue)
     NoMethodError:
       undefined method `post' for nil:NilClass
     # ./lib/logstash/inputs/beats.rb:219:in `start_buffer_broker'
     # ./lib/logstash/inputs/beats.rb:128:in `run'
     # ./spec/inputs/beats_spec.rb:93:in `(root)'

Finished in 24.73 seconds (files took 2.38 seconds to load)
88 examples, 2 failures

Failed examples:

rspec ./spec/inputs/beats_spec.rb:107 # LogStash::Inputs::Beats#handle_new_connection when an exception occur calls flush on the handler and tag the events
rspec ./spec/inputs/beats_spec.rb:19 # LogStash::Inputs::Beats#register raise no exception
```

ph:
@jordansissel I will just refactor the test to go through `#run`

logstash-input-beats.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = "logstash-input-beats"
-  s.version         = "2.0.3"
+  s.version         = "2.0.4"
```
suyograo: Lets bump this to 2.1.0. Its a big overhaul :)

ph: next will be the library itself

ph:
@jsvd I have been trying to reproduce the concurrency error you are having without luck, are you able to reproduce it constantly? its only happening in the shutdown?

jordansissel:
@ph and I chat over zoom and we saw that the stack trace in the report from [a discuss thread](https://discuss.elastic.co/t/problem-with-logstash-and-beats/37376) was using the CircuitBreaker from Lumberjack input _in_ the Beats input -- oops! I'm not sure if this is _the_ cause of the concurrency problem, but it's a bug and @ph will get a fix for it.

jsvd:
@ph when I tested I got the exception consistently at (and only at) shutdown.

ph:
@jsvd Did you generate a lot of events on the filebeat side?

jsvd:
@ph it was reading from a file, and I had a logtash instance doing generator => 10000 to file with a sleep time 1 second, so it was a low cadence

ph:
@jsvd For you error I think its because we are using blocking read on the socket, I force a close on the socket when the shutdown happen. This error look like an `InterruptionException` on the thread that trying to acquire the lock, but in jruby theses errors are re raised as `ConcurrencyError` in that context, since we don't have any message, I suspect this is happening in a lower level of Java.

ph:
@jsvd Okay, the `ConcurrencyError` you were experiencing or other people were I think its due the same thing. I have paired with @colinsurprenant this morning.

The scenario is a bit complex, but lets try to make it simplier:

The server socket is closed, unblocking the thread block on read, this action generate an `InterruptException`, Ruby catches this error and rethrow a `ConcurrencyError`. This error is originating inside the threadpool, which in fact use the Java Executor service to manage the threads.

We don't know why the trace show the mutex, but moving away from the concurrent ruby threadpool and used standard ruby thread solved that issue.

Before using that concurrent primitive we need more understanding of that layer, we might need to be more carefully with our java exception handling in theses scenario.

Can you test/do a final review on your side?

jordansissel:
```
Failures:

  1) LogStash::Inputs::BeatsSupport::DecodedEventTransform map contains a timestamp when its valid uses as the event timestamp
     NoMethodError:
       undefined method `now' for Object:Class
     # ./lib/logstash/inputs/beats_support/circuit_breaker.rb:93:in `state'
     # ./lib/logstash/inputs/beats_support/circuit_breaker.rb:60:in `closed?'
     # ./lib/logstash/inputs/beats.rb:126:in `run'
     # ./spec/inputs/beats_spec.rb:93:in `(root)'

Finished in 26.66 seconds (files took 3.02 seconds to load)
87 examples, 1 failure

Failed examples:

rspec ./spec/inputs/beats_support/decoded_event_transform_spec.rb:47 # LogStash::Inputs::BeatsSupport::DecodedEventTransform map contains a timestamp when its valid uses as the event timestamp

Randomized with seed 36006

```

jordansissel:
Doesn't always reproduce with different random seeds.

ph:
Er, let me try your seed, I think there is a mock that interferer in the run. I removed the other one.

jordansissel:
I reran that seed (36006) another time, and it passed. Funky.

jordansissel:
```
Finished in 26.51 seconds (files took 2.38 seconds to load)
87 examples, 0 failures

Randomized with seed 36006

```

ph:
I think the `expect(Time).to receive(:now)` its not correctly cleaned after a run... I used timecop in the past for theses kinds of trickery.

ph:
Funky, I removed the calls for mocking the Time.
back to the future

guyboertje:
Probably of no use at all but I ran the tests from the original PR checkout many times with no errors. Perhaps its because I am in the UTC timezone or I have a powerful Macbook Pro.

lib/logstash/inputs/beats.rb
```diff
@@ -99,104 +125,82 @@ def run(output_queue)
...
-        copy_beat_hostname(decoded)
-        decorate(decoded)
-        block.call(decoded)
+    # we may have some stuff in the codec buffer
+    transformer = LogStash::Inputs::BeatsSupport::EventTransformCommon.new(self)
+
+    # Go through all the active connection and flush their
+    # codec content, some context data could be lost in this case
+    # but at least the events main data would be persisted.
+    @connections_list.each do |_, v|
```
jsvd: please rename this variable `v` to maybe.. `connection`?

ph: done, renamed connection_handler.

lib/logstash/inputs/beats.rb
```diff
@@ -99,104 +131,82 @@ def run(output_queue)
...
-    @logger.warn("Error parsing @timestamp string, setting current time to @timestamp",
-      :value => ts.inspect, :exception => e.message)
+    @logger.debug("Beats input: stopped")
   end
 
-  private
-  def invoke(connection, &block)
-    @threadpool.post do
+  # This Method is called inside a new thread
+  def handle_new_connection(connection)
```
jsvd: the indentation in this method is strange:

```ruby
def method
    begin
      action()
    rescue => e
      action2()
    ensure
      action3()
    end
end
````
I suggest:
```ruby
def method
  action()
rescue => e
  action2()
ensure
  action3()
end
```

ph: fixed, not a fan of method level rescue.. probably why I never think about using them

lib/logstash/inputs/beats.rb
```diff
@@ -99,104 +131,82 @@ def run(output_queue)
...
       rescue => e # If we have a malformed packet we should handle that so the input doesn't crash completely.
-        @logger.error("Beats input: unhandled exception", :exception => e, :backtrace => e.backtrace)
+        @logger.error("Beats input: unhandled exception", 
+                      :exception => e,
+                      :backtrace => e.backtrace)
+      ensure
+        connection_handler.flush do |event|
+          # handle the basic event enrichment with tags
+          # since at that time we lose all the context
+          LogStash::Inputs::BeatsSupport::EventTransformCommon.new(self).transform(event)
```
jsvd: could we move the creation of the transformer to between line 197 and 198? otherwise we're creating 1 for each event, while it could be reused?

ph: sure same for the `#stop`

lib/logstash/inputs/beats.rb
```diff
@@ -206,9 +210,19 @@ def invoke(connection, &block)
...
+      begin
+        while !stop?
+          @output_queue << @buffered_queue.take
+        end
+      rescue InterruptionException => e
+        # If we are shutting down without waiting the queue to unblock
+        # we will get an `InterruptionException` in that context we will not log it.
+        @logger.error("Beats input: bufferered queue exception", :exception => e) unless stop?
+      rescue => e
+        @logger.error("Beats input: unexpected exception", :exception => e) unless stop?
```
jsvd: have you considered not having "unless stop?" in this branch, as it is an unexpected exception?

ph: Thats a good point, should be ok we only noticed `InterruptionException` coming from that part of the code

jsvd:
LGTM :sheep: it!

PR #40, opened 2016-01-11
Release a new version of the gem without any included dependency

ph:
2.1.0 included the dependencies in the released gem package, this PR
fix the problem and also remove `vendor/` from the gemspec's files
attributes. The 2.1.0 will be yanked from rubygems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.1.1
+  - Release a new version of the gem that doesn't included any other gems, 2.1.0 is yanked from rubygems 
 # 2.1.0
   - Refactor of the code to make it easier to unit test
   - Fix a conncurrency error on high load on the SizeQueue #37
--- a/logstash-input-beats.gemspec
+++ b/logstash-input-beats.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = "logstash-input-beats"
-  s.version         = "2.1.0"
+  s.version         = "2.1.1"
   s.licenses        = ["Apache License (2.0)"]
   s.summary         = "Receive events using the lumberjack protocol."
   s.description     = "This gem is a logstash plugin required to be installed on top of the Logstash core pipeline using $LS_HOME/bin/plugin install gemname. This gem is not a stand-alone program"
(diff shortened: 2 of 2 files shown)


jsvd:
I wonder if the minimum version of logstash-core should be bumped, since clearly it does not work with previous versions due to concurrent ruby clashes. The failed jenkins job shows:

```
...
Using logstash-core 2.0.0
...
Using logstash-input-beats 2.1.1 (was 2.1.0) from source at .
Using bundler 1.10.6
Bundle complete! 8 Gemfile dependencies, 40 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
RuntimeError: Logstash expects concurrent-ruby version 0.9.1 and version 0.9.2 is installed, please verify this patch: /var/lib/jenkins/.rbenv/versions/jruby-1.7.20/lib/ruby/gems/shared/gems/logstash-core-2.0.0-java/lib/logstash/patches/silence_concurrent_ruby_warning.rb
```

ph:
I think the best thing is to leave up the dependency on the core?

ph:
by dependency I meant the concurrent-ruby dep.

jsvd:
I am +1 on bumping the minimum version of logstash-core for this and removing the dependency on concurrent ruby

ph:
i can't remove the dependency but I can let the core handle the constraint?

jsvd:
:+1:

jsvd:
wait, I'm not sure..I like having constraints on all dependencies, specially runtime ones..

shouldn't it pick a logstash-core that also satisfies the concurrent ruby constraint?

ph:
This is the current list of the runtime dependency:

```
  s.add_runtime_dependency "concurrent-ruby", "~> 0.9.2"
  s.add_runtime_dependency "thread_safe", "~> 0.3.5"
  s.add_runtime_dependency "logstash-codec-multiline", "~> 2.0.5"
```

I can replace it with this

```
  s.add_runtime_dependency "concurrent-ruby"
  s.add_runtime_dependency "thread_safe"
  s.add_runtime_dependency "logstash-codec-multiline", "~> 2.0.5"
```

I can't remove the one of the `codec-multiline` since we depend on of the class.

I dont see why we need to depend on a minimum version of core?

WDYT?

ph:
btw, the failing test is because the PR job never clean his workspace between run, because of that the intalled `logstash-core` isnt the latest version. I have changed the job to clean the workspace before actually running the test to make this issue never happen again.

jsvd:
2.1.1 has been published

ph:
I have yanked 2.1.0
```
~ ❯❯❯ gem yank logstash-input-beats --version 2.1.0
Yanking gem from https://rubygems.org...
Successfully deleted gem: logstash-input-beats (2.1.0)
```

PR #41, opened 2016-01-11
Fix a visibility issue accros threads

ph:
When the `#run` save a reference to the actual queue before using the
synchronous queue that reference is not garantee to be visible accross
all the threads. So if you call `#stop` its possible that the value is
nil. So this PR wrap the value inside a `Concurrent::MutexAtomicReference`
and use `#get` on it before inserting new items into the queue.

**Errors on jenkins**
```
the events
     NoMethodError:
     # ./lib/logstash/inputs/beats.rb:193:in `handle_new_connection'
     # ./spec/inputs/beats_spec.rb:112:in `(root)'

  2) LogStash::Inputs::BeatsSupport::ConnectionHandler#process with a codec queue is blocked raise an exception
     NameError:
       uninitialized constant LogStash::Inputs::Beats::InterruptionException
     # ./lib/logstash/inputs/beats.rb:218:in `start_buffer_broker'
```

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -14,6 +14,7 @@ require "logstash/inputs/beats_support/raw_event_transform"
 require "logstash/inputs/beats_support/synchronous_queue_with_offer"
 require "logstash/util"
 require "thread_safe"
+require "concurrent"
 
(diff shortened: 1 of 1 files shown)


ph:
nevermind that not the real fix.

PR #42, opened 2016-01-11
Fix a small issue with the way the tests are setup

ph:
Depending of the load on the CI, this test could work or Not
this PR give a bit more time to start and use Try {} to try multiple
time before failing.

Also rescue `java.lang.InterruptedException` since its a java exception we need to prefix it with the java context.

It should fix the two failing test, which don't happen all the time :(
```
 1) LogStash::Inputs::Beats#handle_new_connection when an exception occur calls flush on the handler and tag the events
     NoMethodError:
     # ./lib/logstash/inputs/beats.rb:193:in `handle_new_connection'
     # ./spec/inputs/beats_spec.rb:112:in `(root)'

  2) LogStash::Inputs::BeatsSupport::ConnectionHandler#process with a codec queue is blocked raise an exception
     NameError:
       uninitialized constant LogStash::Inputs::Beats::InterruptionException
     # ./lib/logstash/inputs/beats.rb:218:in `start_buffer_broker'

```

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -217,9 +217,9 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
         while !stop?
           @output_queue << @buffered_queue.take
         end
-      rescue InterruptionException => e
+      rescue java.lang.InterruptedException => e
         # If we are shutting down without waiting the queue to unblock
-        # we will get an `InterruptionException` in that context we will not log it.
+        # we will get an `java.lang.InterruptionException` in that context we will not log it.
         @logger.error("Beats input: bufferered queue exception", :exception => e) unless stop?
       rescue => e
         @logger.error("Beats input: unexpected exception", :exception => e)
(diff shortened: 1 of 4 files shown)


ph:
for long solution https://github.com/logstash-plugins/logstash-input-beats/issues/43

lib/logstash/inputs/beats.rb
```diff
@@ -217,7 +217,7 @@ def start_buffer_broker
         while !stop?
           @output_queue << @buffered_queue.take
         end
-      rescue InterruptionException => e
+      rescue java.lang.InterruptedException => e
         # If we are shutting down without waiting the queue to unblock
         # we will get an `InterruptionException` in that context we will not log it.
```
jsvd: Please also update the name of the exception here.

Also, is there a way to test this? I'm guessing that putting here ArgumentError would not cause tests to fail

ph: I will update the documentation to make it clear thats not a ruby exception but a java one.

I don't think there is a reliable way to test this, this issues was raised by the current tests, but I never got that locally..


I am not sure if its worth it to use a stub the `#take` call to raise that exception? This seems fragile to me.

jsvd: I'm ok with just updating the docs :+1: tests pass, LGTM

ph:
jenkins test this please

ph:
jenkins test this please

Issue #44, opened 2016-01-15
Beats Input: Remote connection closed unsupported protocol 42 warning


MarkGoldsmith1987:
I'm seeing this filling up my logstash log:

{:timestamp=>"2016-01-15T11:46:37.690000+0000", :message=>"Beats Input: Remote connection closed", :peer=>"xx.xx.xx.xx:41919", :exception=>#<Lumberjack::Beats::Connection::ConnectionClosed: Lumberjack::Beats::Connection::ConnectionClosed wrapping: Lumberjack::Beats::Parser::UnsupportedProtocol, unsupported protocol 42>, :level=>:warn}

Prior to upgrading to the latest version of the logstash-input-beats plugin with was being reported as an error, however since upgrading the logs from filebeat are being processed by its completely spamming the logstash.log file with these warnings.

Versions
Filebeats 1.0.1
Logstash-input-beats  2.1.2
Logstash: 2.1.1

alexkappa:
Same issue here.

```
{:timestamp=>"2016-01-20T11:36:04.761000+0000", :message=>"Beats Input: Remote connection closed", :peer=>"10.244.5.0:39878", :exception=>#<Lumberjack::Beats::Connection::ConnectionClosed: Lumberjack::Beats::Connection::ConnectionClosed wrapping: Lumberjack::Beats::Parser::UnsupportedProtocol, unsupported protocol 72>, :level=>:warn}
```
```
$ filebeat --version
filebeat version 1.0.1 (amd64)
```
```
$ logstash --version
logstash 2.1.1
```

ph:
In the 2.1.0 release of the plugin I have added more granular loggings, but this certainly concerning if you  see this really often in your logs, Is there any correlating log on the filebeat side? 

This error is originating at the protocol level inside the parser state machine, anything else is trying to connect to the 5044 port?

ph:
If you try to connect to the input with telnet and try to send something you will get a protocol error.

```
Settings: Default filter workers: 2
Logstash startup completed
Beats Input: Remote connection closed {:peer=>"0:0:0:0:0:0:0:1:57640", :exception=>#<Lumberjack::Beats::Connection::ConnectionClosed: Lumberjack::Beats::Connection::ConnectionClosed wrapping: Lumberjack::Beats::Parser::UnsupportedProtocol, unsupported protocol 100>, :level=>:warn}
````

@alexkappa  Not sure if its related, the peer address is ending with 0?

[...]

Issue #35, opened 2015-12-17
unhanded exception - unsupported protocol 60


schmandforke:
Hey there,

i got an unusual exception and don't know what todo:

` ` ` 
Beats input: unhandled exception {:exception=>#<RuntimeError: unsupported protocol 60>, :backtrace=>["/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:225:in `handle_version'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:210:in `header'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:163:in `feed'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:342:in `read_socket'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:319:in `run'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/logstash/inputs/beats.rb:184:in `invoke'", "org/jruby/RubyProc.java:281:in `call'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/concurrent-ruby-0.9.2-java/lib/concurrent/executor/executor_service.rb:515:in `run'", "Concurrent$$JavaExecutorService$$Job_463958253.gen:13:in `run'"], :level=>:error}
```

my filbeat config looks like this:

```
filebeat:
  prospectors:
    -
      paths:
        - /var/log/*.log
      input_type: log
output:
  logstash:
    enabled: true
    hosts:
      - logstash01.xxx.com:5522
```

and my logstash config is this:

```
input {
  beats {
    type  => "my_events"
    port  => "5522"
  }
}
filter {
  ...
}
output {
  ...
}
```

would be great if somebody can help me... i don't see what is wrong... 

My Setup is:

elasticsearch (3 nodes) => version 2.1.0-1
logstash (2 nodes) => version 2.1.1-1
kibana ( 1 node) => version 4.3.0
filebeat => version 1.0.0

Thanks

urso:
[...]

Issue #34, opened 2015-12-08
zlib buffer errors


jlintz:
```
{:timestamp=>"2015-12-08T17:20:11.966000-0500", :message=>"Beats input: unhandled exception", :exception=>#<Zlib::BufError: buffer error>, :backtrace=>["org/jruby/ext/zlib/ZStream.java:134:in `finish'", "org/jruby/ext/zlib/JZlibInflate.java:72:in `inflate'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/jls-lumberjack-0.0.26/lib/lumberjack/server.rb:292:in `compressed_payload'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/jls-lumberjack-0.0.26/lib/lumberjack/server.rb:163:in `feed'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/jls-lumberjack-0.0.26/lib/lumberjack/server.rb:296:in `compressed_payload'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/jls-lumberjack-0.0.26/lib/lumberjack/server.rb:163:in `feed'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:342:in `read_socket'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/lumberjack/beats/server.rb:319:in `run'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.0.3/lib/logstash/inputs/beats.rb:184:in `invoke'", "org/jruby/RubyProc.java:271:in `call'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/concurrent-ruby-0.9.2-java/lib/concurrent/executor/executor_service.rb:515:in `run'", "Concurrent$$JavaExecutorService$$Job_1486431032.gen:13:in `run'"], :level=>:error}
```

Seem to be getting this continuously on some nginx logs

Input is just

```
    beats {
      port               => 12348
      codec              => json
    }
```

And the nginx logs are JSON formatted.  I'll see if I can create a reproducible case

Versions:
logstash 2.1.0-1
logstash-input-beats (2.0.3)
filebeat 1.0.0

ph:
@jlintz Thanks! if you can get a lot that would be awesome! @urso Any ideas?

urso:
[...]

PR #47, opened 2016-02-07
Parse compressed body directly

urso:
Fixes connection failures due to buffer handling if data read from socket
contain partial content from next message by parsing decompressed content directly.

Fix is specific for libbeat output plugin and fixes issues encountered in PR elastic/beats#941.
Errors encountered have been similar to: #44, #35, #34

Checking throughput on hetzner test-machine improved by ~10% from 20k eps to 22k eps. Enabling PR elastic/beats#941, throughput increased to ~25.5k eps. With 2 workers it increased to 43k eps.

### DIFF
--- a/lib/lumberjack/beats/server.rb
+++ b/lib/lumberjack/beats/server.rb
@@ -308,7 +308,27 @@ module Lumberjack module Beats
       transition(:header, 2)
 
       # Parse the uncompressed payload.
-      feed(original, &block)
+      # feed(original, &block)
+      off = 0
+      while off < (original.size - 10) # account for data frame header
+        version, code = original[off, 2].bytes.to_a
+        if !(version == PROTOCOL_VERSION_2 and code == FRAME_JSON_DATA)
+          # fallback to original feed method for backwards compatibility to
+          # logstash-forwarder
+          feed(original[off..-1], &block)
+          return
+        end
+
+        off += 2
+        @sequence, payload_size = original[off, 8].unpack('NN')
+        off += 8
+        if (off + payload_size) > original.size
+          raise "JSON payload length exceeds uncompressed message"
+        end
+        payload = original[off, payload_size]
+        off += payload_size
+        yield :json, @sequence, Lumberjack::Beats::json.load(payload)
+      end
     end
   end # class Parser
 


ph:
jenkins test this please

ph:
ooops, I was curious so I run a test coverage on the code.

screenshot 2016-02-09 10 24 59

I think we need to update the client library to be able to use the version 2 of the protocol.
because we dont have any code that trigger theses lines.

ph:
@urso I have written a test for it, will do a bit of testing to see how it behaves.

ph:
@urso, I have compiled the PR from https://github.com/elastic/beats/pull/941
And tested it agains't https://github.com/ph/logstash-input-beats/tree/fix/state-machine-woes using `pipelined` set to true, I think its a bit simplier to just create a new parser if the payload is compressed and reuse the existing code.

WDYT? Did I miss something from your patch?

urso:
Have been thinking about creating a new parser instance first, but found the buffer handling a little tricky in current implementation. The compressed payload contains complete content and parser must not wait for any additional input + must consume complete buffer.

Then parsing the content as being send by beats is quite simple and I found dissecting the buffer as is pretty straight forward (e.g. line 314 to 320 is only for BC).

ph:
@urso Well, creating a new instance and calling the feed with a fixed size should do the same thing as consuming the complete bufffer.

Technically the buffer handling/waiting is external in the while loop? If the performance gain is similar I would prefer that we reuse the state machine and fix the problem with the actual content in the buffer.

ph:
@urso I agree that the current buffer handling is a bit tricky, but I though reusing the state machine make sense in this case.

urso:
Looks to me like buffer handling is kinda embedded into the parse, that's why I did shy away from doing it this way. But I agree, if we can easily re-use the state machine we should do so. Will check if I can get it done this way and if there is any difference in performance.

ph:
@urso I have tried it on https://github.com/ph/logstash-input-beats/blob/fix/state-machine-woes/lib/lumberjack/beats/server.rb#L309-L310 Can you tests it with your setup?

It seen to work fine for me with a filebeat built from https://github.com/elastic/beats/pull/941

matglas:
Is there a nightly build to test with? I commented in #44 and would be able to test this in my Docker setup.

ph:
@matglas We don't do nightly build of the plugins but you can define the gem using with a git source in the logstash's gem file and install it with `bin/plugin install --no-verify`

IIRC, replacing the logstash-input-beats in the Gemfile with this one should work:
```
gem "logstash-input-beats", :git => "https://github.com/ph/logstash-input-beats/" :branch => fix/state-"machine-woes"

ph:
Replaced by https://github.com/logstash-plugins/logstash-input-beats/pull/89

PR #49, opened 2016-02-09
Make sure we stop the thread

### LINKED ISSUE #48: When running the tests inside logstash
This test was failling when we were running all the plugins tests under logtash.
```
1) LogStash::Filters::Date fill next year if january events arrive in december "Jan 01 01:00:00" when processed
     NoMethodError:
       undefined method `now' for Time:Class
     # ./vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.1.2/lib/logstash/inputs/beats_support/circuit_breaker.rb:93:in `state'
     # ./vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.1.2/lib/logstash/inputs/beats_support/circuit_breaker.rb:60:in `closed?'
     # ./vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.1.2/lib/logstash/inputs/beats.rb:126:in `run'
     # ./vendor/bundle/jruby/1.9/gems/logstash-input-beats-2.1.2/spec/inputs/beats_spec.rb:93:in `(root)'
```

ph:
The server was still reading the socket when the test was completed
making other tests fails.

Since we are mocking the connection to test the behavior of thread
connection handler calling `#stop` on the plugin wont correctly shutdown
the plugin. Until we have implemented a non blocking stack for beats
inputs I suggest we force to stop the server at the end of running the
tests.

Note in this change doesnt change the test for making sure the plugin
can be stopped.

Fixes #48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.1.3
+  - Make sure we stop all the threads after running the tests #48
 # 2.1.2
   - Catch the `java.lang.InterruptedException` in the events broker
   - Give a bit more time to the Thread to be started in the test #42
--- a/logstash-input-beats.gemspec
+++ b/logstash-input-beats.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = "logstash-input-beats"
-  s.version         = "2.1.2"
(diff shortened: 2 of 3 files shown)


jsvd:
LGTM

Issue #8, opened 2015-10-27
Verify client certificates against CA


andrenarchy:
Currently, the beats input does not verify the client certificate against the CA. This means that *any* client can connect to *any* logstash server and submit arbitrary data. Besides cluttering the logstash server with meaningless data, attackers could try to take servers down by submitting huge amounts of data.

Analogous to [filebeat](https://github.com/elastic/filebeat)/[logstash-forwarder](https://github.com/elastic/logstash-forwarder), the client certificate should be checked against the CA on the server side.

See also logstash-plugins/logstash-input-lumberjack#31.

klj613:
+1

AndreasWurm:
indeed, stumbled over this too. In Filebeat you can set the accepted CAs and cipher suites.
The same should be true for the receiving end.

bobelev:
+1

runningman84:
This seems to be an important issue, if you want to drop lumberjack in the future you should at least have the same kind of security with beats.

Funny enough filebeat already supports using a client certificate. Logstash filebeat input just fails to validate it.

Please add this option:
```
ssl_client_cert_check
```

jordansissel:
@runningman84 I don't think the lumberjack input does validation (on the logstash input side) of client certificates - the original threat model was to protect data leaks (so lsf doesn't ship to untrusted servers). There's been enough requests for client authentication via certificate that it's worth adding this feature, so  I agree. Hope the background helps.

acchen97:
+1 on implementing this, /cc @suyograo

filipemarinheiro:
+1 For me this is a major security fault and should be implemented as soon as possible

ravi412:
+1 . We are planning a large-scale install of filebeat on 1000+ machines hence this is a critical feature.

ph:
I am starting to look at this following the use cases defined in https://github.com/logstash-plugins/logstash-input-lumberjack/issues/31

rjes:
+1
Since we would like to send log from instances in iaas/paas locations where we can't setup ip-filters this feature is very crucial for us.

acchen97:
Thanks for the feedback, we're targeting this within a short-term release.

erikswanson:
When this gets implemented, can we please have access to information about the client certificate? I want to be able to an event to what client certificate was used to send it to Logstash.

ph:
[...]

PR #56, opened 2016-03-09
TLS Client verification

ph:
This PR introduce client verification when using `ssl`, when using this
option you need to specify the **Certificate Authority** using the
`ssl_certificate_authorities` option, this configuration take a
certificate to validate the client with it. You also need to specify a
`ssl_verify_mode`, If you want strict ssl verification use `force_peer`,
with this value if the client doesn't provide a valid certificate the
plugins will refuse the client.

On Filebeat you need to configure the client to declare a certificate using
theses options.

```
tls:
  certificate_authorities: ["/etc/server.crt"]
  certificate: /etc/server.crt
  certificate_key: /etc/server.key
```

This change also introduce testing with travis-ci and an integration
suite with filebeat for the SSL options.

Fix: #8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2.1.5
+# 2.2.0
   - The server can now do client side verification by providing a list of certificate authorities and configuring the `ssl_verify_mode`,
     the server can use `peer`, if the client send a certificate it will be validated. Using `force_peer` will make sure the client provide a certificate
     and it will be validated with the know CA.  #8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.1.5
+  - The server can now do client side verification by providing a list of certificate authorities and configuring the `ssl_verify_mode`,
+    the server can use `peer`, if the client send a certificate it will be validated. Using `force_peer` will make sure the client provide a certificate
+    and it will be validated with the know CA.  #8
 # 2.1.4
   - Change the `logger#warn` for `logger.debug` when a peer get disconnected, keep alive check from proxy can generate a lot of logs  #46
 # 2.1.3
--- a/lib/lumberjack/beats/server.rb
+++ b/lib/lumberjack/beats/server.rb
@@ -29,35 +29,33 @@ module Lumberjack module Beats
         :ssl => true,
         :ssl_certificate => nil,
         :ssl_key => nil,
-        :ssl_key_passphrase => nil
+        :ssl_key_passphrase => nil,
+        :ssl_certificate_authorities => nil,
+        :ssl_verify_mode => :none # By default we dont verify client
       }.merge(options)
 
       if @options[:ssl]
-        [:ssl_certificate, :ssl_key].each do |k|
-          if @options[k].nil?
-            raise "You must specify #{k} in Lumberjack::Server.new(...)"
-          end
+        if verify_client?(@options[:ssl_verify_mode]) && certificate_authorities.empty?
+          raise "When `ssl_verify_mode` is set to `peer` OR `force_peer` you need to specify the `ssl_certificate_authorities`"
+        end
+
+        if !verify_client?(@options[:ssl_verify_mode]) && certificate_authorities.size > 0 
+          raise "When `ssl_certificate_authorities` is configured you need to set `ssl_verify_mode` to either `peer` or `force_peer`"
+        end
+
+        if @options[:ssl_certificate].nil? || @options[:ssl_key].nil?
+          raise "You must specify `ssl_certificate` AND `ssl_key`"
         end
       end
 
       @server = TCPServer.new(@options[:address], @options[:port])
-
       @close = Concurrent::AtomicBoolean.new
+      @port = retrieve_current_port
 
-      # Query the port in case the port number is '0'
-      # TCPServer#addr == [ address_family, port, address, address ]
-      @port = @server.addr[1]
-
-      if @options[:ssl]
-        # load SSL certificate
-        @ssl = OpenSSL::SSL::SSLContext.new
-        @ssl.cert = OpenSSL::X509::Certificate.new(File.read(@options[:ssl_certificate]))
-        @ssl.key = OpenSSL::PKey::RSA.new(File.read(@options[:ssl_key]),
-          @options[:ssl_key_passphrase])
-      end
+      setup_ssl if ssl?
     end # def initialize
 
-    # Server#run method, allow the library to manage all the connection 
+    # Server#run method, allow the library to manage all the connection
     # threads, this handing is quite minimal and don't handler
     # all the possible cases deconnection/connection.
     #
@@ -76,8 +74,8 @@ module Lumberjack module Beats
         Thread.new(connection) do |connection|
           begin
             connection.run(&block)
-          rescue Lumberjack::Beats::Connection::ConnectionClosed 
-            # Connection will raise a wrapped exception upstream, 
+          rescue Lumberjack::Beats::Connection::ConnectionClosed
+            # Connection will raise a wrapped exception upstream,
             # but if the threads are managed by the library we can simply ignore it.
             #
             # Note: This follow the previous behavior of the perfect silence.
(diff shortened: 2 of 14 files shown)


ph:
@jordansissel Can you take a look on this? I have small issue with intermediate certificate can we pair on this?

spec/support/integration_shared_context.rb
```diff
@@ -0,0 +1,70 @@
...
+    execute_client
+  end
+
+  it "successfully send the events" do
+    expect(queue.size).to eq(number_of_events), "Expected: #{number_of_events} got: #{queue.size}, execution output:\n #{@execution_output}"
+    expect(queue.collect { |e| e["message"] }).to eq(events)
+  end
+end
+
+shared_examples "doesn't events" do
```
ph: should be `doesn't send eventd`. will fix it.

lib/lumberjack/beats/server.rb
```diff
@@ -76,8 +67,8 @@ def run(&block)
         Thread.new(connection) do |connection|
           begin
             connection.run(&block)
-          rescue Lumberjack::Beats::Connection::ConnectionClosed 
-            # Connection will raise a wrapped exception upstream, 
+          rescue Lumberjack::Beats::Connection::ConnectionClosed => e
```
jordansissel: (Low priority, style) Any reason to capture the exception here when we aren't using it?

lib/lumberjack/beats/server.rb
```diff
@@ -139,13 +130,80 @@ def close
...
+      # TCPServer#addr == [ address_family, port, address, address ]
+      @server.addr[1]
+    end
+
+    def certificate_authorities
+      Array(@options[:ssl_certificate_authorities])
+    end
+
+    def server_private_key
+      OpenSSL::PKey::RSA.new(File.read(@options[:ssl_key]),
```
jordansissel: So this method will return nil if `ssl_key` is not provided? Should it be an exception if `server_private_key` is invoked but `:ssl_key` is not provided?

lib/lumberjack/beats/server.rb
```diff
@@ -139,13 +130,80 @@ def close
...
+                             @options[:ssl_key_passphrase]) if @options[:ssl_key]
+    end
+
+    def server_certificate
+      OpenSSL::X509::Certificate.new(File.read(@options[:ssl_certificate])) if @options[:ssl_certificate]
```
jordansissel: Same comment here; should we raise exception if there is no `:ssl_certificate` option given, instead of returning nil?

lib/lumberjack/beats/server.rb
```diff
@@ -139,13 +130,80 @@ def close
...
+
+      if certificate_authorities.size > 0
+        certificate_authorities.each do |certificate_authority|
+          if File.file?(certificate_authority)
+            store.add_file(certificate_authority)
+          else
+            # `#add_path` is not implemented under jruby
+            # so recursively try to load all the certificate from this directory
+            # https://github.com/jruby/jruby-openssl/blob/master/src/main/java/org/jruby/ext/openssl/X509Store.java#L159
+            if !!(RUBY_PLATFORM == "java")
```
jordansissel: why use `!!` here?

spec/integration/filebeat_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+        "scan_frequency" => "1s"
+      },
+      "output" => { 
+        "logstash" => { "hosts" => ["#{host}:#{port}"] },
+        "logging" => { "level" => "debug" }
+      }
+    }
+  end
+
+  let_tmp_file(:filebeat_config_path) { YAML.dump(filebeat_config) }
```
jordansissel: Love this let_tmp_file thing.

spec/integration_spec.rb
```diff
@@ -10,23 +10,39 @@
 Thread.abort_on_exception = true
 describe "A client" do
   let(:certificate) { Flores::PKI.generate }
-  let(:certificate_file_crt) { "certificate.crt" }
-  let(:certificate_file_key) { "certificate.key" }
+  let(:certificate_file_crt) do
```
jordansissel: use `let_tmp_file` here?

jordansissel:
Overall this is awesome. I left a few style comments which you are free to address now or in a future PR.

Specs pass for me on my laptop.

LGTM w/ some style comments to resolve if you wish.

lib/lumberjack/beats/server.rb
ph: Ooops A debug trace, I will remove the => e

spec/support/flores_extensions.rb
```diff
@@ -0,0 +1,42 @@
+# encoding: utf-8
+require "flores/pki"
+module Flores
+  module PKI
```
jordansissel: +1 on moving this into Flores library.

spec/integration_spec.rb
ph: Yes, I have created `let_tmp_file` after that spec file, it was pissing me off to repeat the code :P

but yes will replace all theses by it, since it makes the code more readable.

harbulot:
Just wondering if it might be worth considering harmonising the settings across the various plugins, possibly related to logstash-plugins/logstash-input-tcp#9 (which should be re-opened) and logstash-plugins/logstash-input-tcp#42 to some extent.

jordansissel:
@harbulot it would be nice to have the same setting names-and-meanings for shared concepts (x509 certs, etc) among plugins, yes. Can you open a new issue to discuss that?

ph:
@harbulot Yes this is on the plan, still havent read completly all your latest comments concerning ssl issues on the tcp input.

lib/lumberjack/beats/server.rb
ph: fixed

lib/lumberjack/beats/server.rb
ph: Good point I think the validation in the `#initialize` is wrong I will reflect to make the following change.

When `ssl` is set to true:
- **ssl_key** need to be set.
- **ssl_certificate** need to be set.

When `ssl_verify_mode` is set to other than `none`:
- **ssl_certificate_authorities** need to be set.

Its doesnt make sense to me to be able to remove the certificate and the key if we turn on ssl.

spec/support/integration_shared_context.rb
ph: done

lib/lumberjack/beats/server.rb
ph: See my previous note concerning validation of the options.
https://github.com/logstash-plugins/logstash-input-beats/pull/56#discussion_r56160334

lib/lumberjack/beats/server.rb
ph: should not be needed.

harbulot:
@jordansissel Sure, in which project? `logstash-input-beats` or `-tcp`? I would have thought this would nicely fit with logstash-plugins/logstash-input-tcp#42 (since harmonisation and pre-validation of the parameters would certainly go well together). I'm not against creating another SSL-related issue, but I guess splitting everything into sub-issues can be a factor that leads to this heterogeneity in the settings.

ph:
@jordansissel I have created:

```
https://github.com/jordansissel/ruby-flores/issues/6
https://github.com/jordansissel/ruby-flores/issues/5
```
For future refactor, I will do it when I go back to the tcp input.

ph:
@harbulot I think the issue belongs into core since we have a lot of plugins that can handle some kind of ssl option I think its fair to have a common interface.

Lets move the discussion to https://github.com/elastic/logstash/issues/2151

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.1.5
```
ph: @jordansissel Should it be 2.2.0, since we are introducing new options?

ph:
@jordansissel Updated with most of you changes, minus the extraction of the flores and rspec helpers.
What version we should use for that ? 2.1.5 or 2.2.0 ?

jordansissel:
@ph +1 to using 2.2.0 since we're adding new settings. We can do the flores+rspec extraction in another PR.

Once version is updated to 2.2.0, LGTM is available :)

PR #59, opened 2016-03-18
New version constraints for the Logstash 5.0 release

ph:


ph:
Also updated the changelog..

logstash-input-beats.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.metadata = { "logstash_plugin" => "true", "logstash_group" => "input" }
 
   # Gem dependencies
-  s.add_runtime_dependency "logstash-core", ">= 2.0.0", "< 3.0.0"
+  s.add_runtime_dependency "logstash-core", ">= 2.0.0", "< 6.0.0"
```
jsvd: we need to ensure alpha1 here. We don't know if 6.0 will break backwards compatibility. Keep it as is means that a logstash beta2 deployment of 6.0.0 can still install these plugins.

ph: true so this should be 6.0.0.alpha1 instead of 6.0.0

ph:
update with the files generated from the scripts.

jsvd:
LGTM

ph:
Part of a mass update done March 18 2016 2h30PM EST

PR #65, opened 2016-03-25
Make `ssl_key_passphrase` work

### LINKED ISSUE #61: SSL_key_passphrase doesn't currently work.
Discovered when debugging an issue in https://github.com/logstash-plugins/logstash-input-tcp/issues/48
We need to call `.value` here to make sure we have the right key to decrypt the private key

https://github.com/logstash-plugins/logstash-input-beats/blob/master/lib/logstash/inputs/beats.rb#L118

suyograo:
Fix bug where using `ssl_key_passphrase` wouldn't work

Fixes #61


talevy:
LGTM

Issue #66, opened 2016-03-28
IdentityMapCodec has reached 100% capacity


hrchu:
logstash 2.2.2
filebeat 1.1.2
arch: filebeat=>logstash=>elasticsearch

logstash continually output the following error message, and log cannot be write to elasticsearch.

```
{:timestamp=>"2016-03-28T13:48:10.861000+0800", :message=>"IdentityMapCodec has reached 100% capacity", :current_size=>20000, :upper_limit=>20000, :level=>:error}
[...]

PR #70, opened 2016-03-30
Do not use the identity map unless we specify the multiline codec.

ph:
In some case when multiples FB clients are connected to the server and
and watching a lot of files we can hit the limit or the 20K identity
stream. This PR doesnt solve the 20K problems, but changes when the
identity map is used, we will only use it if the multiline codec is
defined in the configuration file. When its not the case the codec will
be reuse accross connection thread. We can do that since our codec are
stateless.

Help with #66 when not using the multiline codec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.2.6
+  - Do not use the identity map if we don't explicitly use the `multiline` codec
 # 2.2.5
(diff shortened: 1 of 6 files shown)


ph:
@ESamir @guyboertje The first we talk earlier this morning.

guyboertje:
LGTM - (did not test locally)

PR #71, opened 2016-03-30
Fix non deterministic tests with ports

### LINKED ISSUE #60: Test failling because the port was already in use.
Port already in use on jenkins, I am pretty sure we use random port on each examples.

http://build-eu-00.elastic.co/job/logstash-plugin-input-beats-unit/jdk=JDK8,nodes=metal-pool/12/console

```
  1) A client using plain tcp connection When transmitting a payload v1 frame when sequence start at 0 supports single element
     Failure/Error: ssl_server = Lumberjack::Beats::Server.new(config_ssl)
     Errno::EADDRINUSE:
       Address already in use - bind - Address already in use
     # ./lib/lumberjack/beats/server.rb:51:in `initialize'
     # ./spec/integration_spec.rb:49:in `(root)'

  2) A client Mutual validation with certificate authorities directly signing a certificate when sequence start at 0 support sending multiple elements in one payload
     Failure/Error: tcp_server = Lumberjack::Beats::Server.new(config_tcp)
     Errno::EADDRINUSE:
       Address already in use - bind - Address already in use
     # ./lib/lumberjack/beats/server.rb:51:in `initialize'
     # ./spec/integration_spec.rb:48:in `(root)'

  3) A client Mutual validation with certificate authorities Mutiple CA different clients when the CA are defined individually in the configuration client from secondary CA when sequence doesn't start at zero supports single element
     Failure/Error: tcp_server = Lumberjack::Beats::Server.new(config_tcp)
     Errno::EADDRINUSE:
       Address already in use - bind - Address already in use
     # ./lib/lumberjack/beats/server.rb:51:in `initialize'
     # ./spec/integration_spec.rb:48:in `(root)'

```

ph:
Instead of just using a random number from a range, The
`Flores::Random.port` will try to setup a server on the specific port,
if everything is fine it will close the server and give you the port
number to use and we will know for sure that the port can be used.
The helper will try multiple port before giving up completely. (15
times)

This PR also make sure that all the tests are using random ports.

fixes #60

### DIFF
--- a/spec/support/flores_extensions.rb
+++ b/spec/support/flores_extensions.rb
@@ -1,6 +1,46 @@
 # encoding: utf-8
 require "flores/pki"
+require "flores/random"
+
 module Flores
+  module Random
+    DEFAULT_PORT_RANGE = 1024..65535
+    DEFAULT_PORT_CHECK_TIMEOUT = 1
+    DEFAULT_MAXIMUM_PORT_FIND_TRY = 15
+
+    class << self
+      def port(range = DEFAULT_PORT_RANGE)
+        try = 0
+        while try < DEFAULT_MAXIMUM_PORT_FIND_TRY
+          candidate = integer(range)
+
+          if port_available?(candidate)
+            break
+          else
+            try += 1
+          end
+        end
+        
+        raise "Flores.random_port: Cannot find an available port, tried #{DEFAULT_MAXIMUM_PORT_FIND_TRY} times, range was: #{range}" if try == DEFAULT_MAXIMUM_PORT_FIND_TRY
+
+        candidate
+      end
+      
+      def port_available?(port)
+        begin
+          server = TCPServer.new(port)
+          available = true
+        rescue # Assume that any errors can do this
+          available = false
+        ensure
+          server.close if server
+        end
+
+        return available
+      end
+    end
+  end
+
   module PKI
     DEFAULT_CERTIFICATE_OPTIONS = {
       :duration => Flores::Random.number(100..2000),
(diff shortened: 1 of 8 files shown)


ph:
Fixes theses http://build-eu-00.elastic.co/job/logstash-plugin-input-beats-unit/22/

spec/support/flores_extensions.rb
```diff
@@ -1,6 +1,46 @@
 # encoding: utf-8
 require "flores/pki"
+require "flores/random"
+
 module Flores
```
andrewvc: This is an awesome util! I assume this was done as a monkey patch due to @jordansissel being out of pocket for a while?

Have you considered just putting this in some other namespace for now? Monkey patching is kind of dicey. Ideally this would just be in `BeatsInputTestHelpers` or the like, and we'd open a PR agains Flores to add this in, then swap it out if that was accepted.

ph: Well, I do this for 3 reasons.. 
- input beats is kind of weird beast to test and I have to create a lot of helpers for that..
- It easier to test in an actual plugins than having to manager multiples gems revision ;)
- I have access to rubygems for flores, but not on the github project itself so I cannot merge on it. (not sure if jarvis has access to it though, I haven't tested it.)

For monkey patch, I think its OK since we are controlling the flores library and its only for testing purposes, I will move it ASAP :tm: 

Well if you like this helper you should look for `let_tmp_file(:lol) { "Hello world" }` in https://github.com/logstash-plugins/logstash-input-beats/blob/master/spec/support/file_helpers.rb

andrewvc: :+1:

andrewvc:
LGTM

PR #72, opened 2016-03-30
LSF Integration test

### LINKED ISSUE #52: Add integration test for LSF to beats input
Discussed with @jordansissel earlier.  He likes the idea of officially supporting LSF -> LS beats input to ease migration.  He has tested this briefly before and it works because LSF and beats use the same underlying protocol, but would like to see us add integration tests so that we can officially confirm and document it as a supported path.

ph:
This changes add integration test with the `Logstash-Forwarder`

Fixes #52

depends on #71


spec/support/flores_extensions.rb
```diff
@@ -1,6 +1,46 @@
...
+          available = false
+        ensure
+          server.close if server
+        end
+
+        return available
+      end
+    end
+  end
+
```
ph: @andrewvc I will make a PR later to move that to flores itself, I have created customs helpers here in the beats input that would add more feature to the flores library.

andrewvc:
LGTM

PR #76, opened 2016-04-12
Fix/lsf integration fails

ph:
This was a nasty bug, I did not have it all the time, It might be related to the jruby version I was running at the time I created the tests.

The problem was the following, LSF metadata doesn't have the right encoding and was making the server thread crash. You can see it by enabling `DEBUG` when running the integration test.

This PR also make sure the plugin is a drop in replacement for lumberjack, since the hash send by LSF and filebeat are a bit different.

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -95,7 +95,7 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   config :congestion_threshold, :validate => :number, :default => 5
 
   # This is the default field that the specified codec will be applied
-  config :target_field_for_codec, :validate => :string, :default => "message"
+  config :target_field_for_codec, :validate => :string, :default => "message", :deprecated => "This option is now deprecated, the plugin is now compatible with Filebeat and Logstash-Forwarder"
 
   # TODO(sissel): Add CA to authenticate clients with.
   RECONNECT_BACKOFF_SLEEP = 0.5
--- a/lib/logstash/inputs/beats_support/connection_handler.rb
+++ b/lib/logstash/inputs/beats_support/connection_handler.rb
@@ -2,6 +2,7 @@
 require "logstash/inputs/beats"
 require "logstash/inputs/beats_support/decoded_event_transform"
 require "logstash/inputs/beats_support/raw_event_transform"
+require "lumberjack/beats"
 
 module LogStash::Inputs::BeatsSupport
   # Handle the data coming from a connection
@@ -40,7 +41,11 @@ module LogStash::Inputs::BeatsSupport
                                       :peer => @connection.peer)
 
       # Filebeats uses the `message` key and LSF `line`
-      target_field = @input.target_field_for_codec ? hash.delete(@input.target_field_for_codec) : nil
+      target_field = if from_filebeat?(hash)
+                       hash.delete(Lumberjack::Beats::FILEBEAT_LOG_LINE_FIELD)
+                     elsif from_logstash_forwarder?(hash)
+                       hash.delete(Lumberjack::Beats::LSF_LOG_LINE_FIELD)
+                     end
 
       if target_field.nil?
         @logger.debug? && @logger.debug("Beats input: not using the codec for this event, can't find the codec target field",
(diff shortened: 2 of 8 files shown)


lib/logstash/inputs/beats.rb
```diff
@@ -95,7 +95,7 @@ class InsertingToQueueTakeTooLong < Exception; end
   config :congestion_threshold, :validate => :number, :default => 5
 
   # This is the default field that the specified codec will be applied
-  config :target_field_for_codec, :validate => :string, :default => "message"
+  config :target_field_for_codec, :validate => :string, :default => "message", :deperecated => "This option is now deprecated, the plugin is now compatible with Filebeat and Logstash-Forwarder"
```
jsvd: deperecated => deprecated

lib/logstash/inputs/beats_support/connection_handler.rb
```diff
@@ -40,7 +43,11 @@ def process(hash, identity_stream)
                                       :peer => @connection.peer)
 
       # Filebeats uses the `message` key and LSF `line`
-      target_field = @input.target_field_for_codec ? hash.delete(@input.target_field_for_codec) : nil
+      target_field = if from_filebeat?(hash)
+                       hash.delete(FILEBEAT_LOG_LINE_FIELD)
+                     elsif from_logstash_forwarder?(hash)
+                       hash.delete(LSF_LOG_LINE_FIELD)
```
jsvd: should we worry about `else` ?

lib/logstash/inputs/beats_support/connection_handler.rb
```diff
@@ -75,5 +82,14 @@ def flush(&block)
 
       @codec.flush(&block)
     end
+
+    private
+    def from_filebeat?(hash)
+      !hash[FILEBEAT_LOG_LINE_FIELD].nil?
```
jsvd: any reason not to do just `hash[FILEBEAT_LOG_LINE_FIELD]`?

lib/lumberjack/beats/server.rb
```diff
@@ -485,6 +485,14 @@ def read_socket(&block)
       end
     end
 
+    def normalize_v1_metadata_encoding(map)
+      # lets normalize the metadata of the v1 frame to make
+      # sure everything is utf-8 minus the content itself.
+      # We need this to make this plugin backward compatible.
```
jsvd: can you elaborate more on why this is needed? why is the content special in that it can't be force encoded too?

lib/lumberjack/beats/server.rb
```diff
@@ -485,6 +485,14 @@ def read_socket(&block)
...
+      map.each { |k, v| map[k].force_encoding(Encoding::UTF_8) unless k == "line" }
```
jsvd: I suggest using `LSF_LOG_LINE_FIELD` here

lib/logstash/inputs/beats_support/connection_handler.rb
ph: Since there is json involved I was worried about receiving a fixnum.

lib/logstash/inputs/beats_support/connection_handler.rb
ph: The other if will take care of it, it means we have received data from something other than filebeat itself.. Could be `topbeat`

lib/lumberjack/beats/server.rb
ph: Oh true, I will put the constant here.

lib/lumberjack/beats/server.rb
ph: LSF is a weird beast, it doesn't return the metadata content in utf-8 (path.. offset) etc and this make logstash crash.
If you look at the lumberjack input https://github.com/logstash-plugins/logstash-input-lumberjack/blob/master/lib/logstash/inputs/lumberjack.rb#L112, I though it was best to move it closer to the actual data parsing. At that stage we know its from LSF or filebeat.
On the other hand filebeat is good citizen :(

ph:
@jsvd I have applied your review, anything else?

guyboertje:
@jsvd - thanks for doing the review.

jsvd:
LGTM, confirmed using this branch allowed logstash to correctly receive data from beats/lumberjack output/lsf

PR #77, opened 2016-04-19
Fix/testing slack token

ph:
Nothing to see here.


ph:
was used for testing the slack integration with any token.

PR #81, opened 2016-04-27
trying to use a checkout of logstash

ph:


ph:
This is an experiments that I have made with @colinsurprenant  to test if it was possible to make a travis job use the unreleased version of the core and it work if the jar is included or correctly builded.

PR #86, opened 2016-05-11
Relax the dependencies on `logstash-devutils`

ph:
The right version for logstash-devutils will be selected on the
logstash-core-plugin-api constraints

Related to https://github.com/elastic/logstash-devutils/issues/48


colinsurprenant:
LGTM

PR #93, opened 2016-06-09
[WIP] Feature/java implementation

ph:
Not ready for production, do not use.


ph:
@talevy there is a faillings test in when the when the client authenticate with a secondary ca, I am looking at fixing this but You can do a first run :)

ph:
@talevy this PR need to run with logstash master.

Gemfile
```diff
@@ -2,3 +2,4 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in logstash-mass_effect.gemspec
 gemspec
+gem "logstash-core", :path => "~/es/logstash"
```
talevy: leaving a note here to remember we need this to go before merging

beats.ipr
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
talevy: git rm this file

beats.iws
```diff
@@ -0,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
talevy: git rm this file

lib/logstash/inputs/beats.rb
```diff
@@ -152,118 +164,28 @@ def target_codec_on_field?
...
-        Thread.new do
-          handle_new_connection(connection)
-        end
-      else
-        @logger.warn("Beats input: the pipeline is blocked, temporary refusing new connection.",
-                     :reconnect_backoff_sleep => RECONNECT_BACKOFF_SLEEP)
-        sleep(RECONNECT_BACKOFF_SLEEP)
-      end
-    end
+    message_listener = MessageListener.new(output_queue, self)
```
talevy: I think it would be better to pass what is needed from `self` here instead of all of `self`.

log4j2.xml_bk
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
talevy: `.xml_bk`?

settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'beats'
```
talevy: I don't think we need this since there is only one project, I've found this file to be helpful for when there are submodules. what do you think? I am fine either way, mostly probing to hear opinion

src/main/java/org/logstash/beats/Batch.java
```diff
@@ -0,0 +1,47 @@
+package org.logstash.beats;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Batch {
+    private byte protocol = Protocol.VERSION_2;
+    private int windowSize;
```
talevy: I think we should keep consistent naming around `Batch` and `Window`. We seem to say that Batches have a window size. Since windows are commonly used to describe overlapping batches, I think I'd prefer the term Batch over Window. so maybe we can rename `windowSize` to `batchSize`?

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+            if(e.state() == IdleState.WRITER_IDLE) {
+                this.sendKeepAlive();
+            } else if(e.state() == IdleState.READER_IDLE) {
+                this.clientTimeout();
+            }
+        }
+    }
+
+    private boolean needAck(Message message) {
```
talevy: how about

```java
private boolean needAck(Message message) {
    return message.getSequence() == message.getBatch().getWindowSize()
}
```

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,225 @@
...
+                    this.batchComplete();
+                }
+
+                transitionToReadHeader();
+                break;
+            }
+        }
+    }
+
+    private boolean hasEnoughBytes(ByteBuf in) {
```
talevy: shorthand

```java
private boolean hasEnoughBytes(ByteBuf in) {
    in.readableBytes() >= this.requiredBytes
}
```

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,225 @@
...
+        transition(States.READ_HEADER, 1);
+    }
+
+    public void transition(States next, long need) {
+        logger.debug("Transition, from: " + this.currentState + " to: " + next + " required bytes: " + need);
+        this.currentState = next;
+        this.requiredBytes = need;
+    }
+
+    public void batchComplete() {
```
talevy: I think this should be private

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,225 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    public void transitionToReadHeader() {
+        transition(States.READ_HEADER, 1);
+    }
+
+    public void transition(States next, long need) {
```
talevy: I think this should be private

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    private boolean hasEnoughBytes(ByteBuf in) {
+        if(in.readableBytes() >= this.requiredBytes) {
+            return true;
+        }
+        return false;
+    }
+
+    public void transitionToReadHeader() {
```
talevy: I think this should be private

src/main/java/org/logstash/beats/IMessageListener.java
```diff
@@ -0,0 +1,13 @@
+package org.logstash.beats;
+
+import io.netty.channel.ChannelHandlerContext;
+
+/**
+ * Created by ph on 2016-05-18.
+ */
+public interface IMessageListener {
+    public void onNewMessage(ChannelHandlerContext ctx, Message message);
```
talevy: can we add javadocs here to explain the purpose of each of these?

src/main/java/org/logstash/beats/Message.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Map;
+
+public class Message implements Comparable<Message> {
+    private int sequence;
+    private String identityStream;
+    private Map data;
+    private Batch batch;
+
+    public Message(int sequence, Map map) {
+        setSequence(sequence);
```
talevy: I think we can do without `setSequence` and `setData` and just set them directly in the constructor

src/main/java/org/logstash/beats/Message.java
```diff
@@ -0,0 +1,64 @@
+package org.logstash.beats;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Message implements Comparable<Message> {
+    private int sequence;
```
talevy: all these fields should be final

src/main/java/org/logstash/beats/Message.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public Batch getBatch() {
+        return batch;
+    }
+
+    public void setBatch(Batch newBatch) {
+        batch = newBatch;
+    }
+
+    public String getIdentityStream() {
```
talevy: should this be done in the constructor?

src/main/java/org/logstash/beats/MessageListener.java
```diff
@@ -0,0 +1,29 @@
+package org.logstash.beats;
+
+import io.netty.channel.ChannelHandlerContext;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+
+// This need to be implemented in Ruby
+public class MessageListener implements IMessageListener {
```
talevy: does jruby not do well with interfaces? and you include this instead for monkeypatching later in jruby via "include" and re-defining?

src/main/resources/log4j.properties
```diff
@@ -0,0 +1,7 @@
+# This sets the global logging level and specifies the appenders
```
talevy: since this is meant to be a library, should we only load this file into the classpath when building the Runner executable jar? and leave it out of the library jar

src/test/java/org/logstash/netty/PrivateKeyConverterTest.java
```diff
@@ -0,0 +1,7 @@
+package org.logstash.netty;
+
+/**
+ * Created by ph on 2016-06-07.
+ */
+public class PrivateKeyConverterTest {
```
talevy: wip?

src/main/java/org/logstash/beats/MessageListener.java
ph: JRuby does fine with interface, you just need to include it?
I dont see any specific issue here.

src/main/resources/log4j.properties
ph: I will remove this file, actually I used locally the .xml config.

src/test/java/org/logstash/netty/PrivateKeyConverterTest.java
ph: oops!

src/main/java/org/logstash/beats/Message.java
ph: +1 I will move it there.

src/main/java/org/logstash/beats/Batch.java
ph: Agree with your naming suggestion, Window doesnt fit the bill for me.

settings.gradle
ph: I dont have a strong opinion on this file, its was created by intellij when scaffolding a new gradle project. So I can remove it

log4j2.xml_bk
ph: this should be in the ignore file

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -0,0 +1,109 @@
...
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@ChannelHandler.Sharable
+public class BeatsHandler extends ChannelInboundHandlerAdapter {
```
spinscale: if you use `SimpleChannelInboundHandler`, you can use generics and dont need to cast in `channelRead`

lib/logstash/inputs/beats.rb
ph: I can't remove the `self` for now because we need to have access to the plugin to be able to decorate the event with the correct tags or types.

```
    def transform(event)
      copy_beat_hostname(event)
      decorate(event)
      event
    end
```

ph:
@talevy Can you take a second look?
@spinscale I have upgrade to netty 4.1.1, I had a few issues with the upgrade that the BeatsParser test catched and I have fixed all of them. I am still doing some test locally for freeing up the bytebuffer to make sure It doesnt have a leak.

build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.hamcrest:hamcrest-library:1.3'
+
+    compile 'org.bouncycastle:bcprov-jdk15on:1.54'
+    compile 'org.bouncycastle:bcpkix-jdk15on:1.54'
+    compile 'io.netty:netty-all:4.1.1.Final'
+    compile 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork17'
+    compile 'org.apache.logging.log4j:log4j-api:2.6.1'
+    compile 'org.apache.logging.log4j:log4j-core:2.6.1'
+    compile 'org.lucee:javassist:3.9.0.GA'
```
spinscale: thats not the origininal javassist package, try http://mvnrepository.com/artifact/org.javassist/javassist

build.gradle
```diff
@@ -0,0 +1,55 @@
...
+}
+
+dependencies {
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.hamcrest:hamcrest-library:1.3'
+
+    compile 'org.bouncycastle:bcprov-jdk15on:1.54'
+    compile 'org.bouncycastle:bcpkix-jdk15on:1.54'
+    compile 'io.netty:netty-all:4.1.1.Final'
+    compile 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork17'
```
spinscale: is this cross platform? Looks like it, but might be worth a test

build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    compile 'org.apache.logging.log4j:log4j-api:2.6.1'
+    compile 'org.apache.logging.log4j:log4j-core:2.6.1'
+    compile 'org.lucee:javassist:3.9.0.GA'
+    compile 'com.fasterxml.jackson.core:jackson-core:2.7.4'
+    compile 'com.fasterxml.jackson.core:jackson-annotations:2.7.4'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.7.4'
+    compile 'com.fasterxml.jackson.module:jackson-module-afterburner:2.7.4'
```
spinscale: 2.7.5 is uptodate, I think the important part is not to have other plugins with different versions in your classpath

src/main/java/org/logstash/beats/Protocol.java
```diff
@@ -0,0 +1,22 @@
+package org.logstash.beats;
+
+/**
+ * Created by ph on 2016-05-16.
+ */
+public class Protocol {
+    public static final byte VERSION_1 = '1';
```
spinscale: you could create an enum out of the protocols, so you dont have to handle bytes but protocol versions instead

src/main/java/org/logstash/beats/BatchEncoder.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    protected void encode(ChannelHandlerContext ctx, Batch batch, ByteBuf out) throws Exception {
+        out.writeByte(batch.getProtocol());
+        out.writeByte('W');
+        out.writeInt(batch.size());
+        out.writeBytes(getPayload(ctx, batch));
+    }
+
+    protected ByteBuf getPayload(ChannelHandlerContext ctx, Batch batch) throws IOException {
+        ByteBuf payload = ctx.alloc().buffer();
```
spinscale: and here begin the nitty gritty details. By default (please use a debugger or print out `payload.isDirect()` to verify) this allocates a direct pooled byte buffer in netty 4.1 - which means it is outside the JVM and not in GC reach (and by default it is unbounded IIRC). So this requires some monitoring during testing. You can use `-XX:MaxDirectMemorySize` as jvm start up arg to limit this. Also visualvm allows you to install a `buffer pools` plugin which monitors direct memory

also, if you know the size of the batch, you can allocate the buffer into the right size

src/main/java/org/logstash/beats/Message.java
```diff
@@ -0,0 +1,61 @@
+package org.logstash.beats;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class Message implements Comparable<Message> {
+    private final int sequence;
+    private String identityStream;
+    private final Map data;
```
spinscale: can you keep this as `Map<String, String>` - prevents casting in the MessageEncoder

src/main/java/org/logstash/beats/BatchEncoder.java
```diff
@@ -0,0 +1,69 @@
...
+        payload.writeBytes(json);
+    }
+
+    protected void encodeMessageWithFields(ByteBuf payload, Message message) {
+        payload.writeByte(Protocol.VERSION_1);
+        payload.writeByte('D');
+        payload.writeInt(message.getSequence());
+        payload.writeInt(message.getData().size());
+
+        for(Object entry : message.getData().entrySet()) {
```
spinscale: This could become `Entry<String, String> entry` to prevent casts?

src/main/java/org/logstash/beats/BatchEncoder.java
```diff
@@ -0,0 +1,69 @@
...
+            Map.Entry e = (Map.Entry) entry;
+            byte[] key = ((String) e.getKey()).getBytes();
+            byte[] value = ((String) e.getValue()).getBytes();
+
+            logger.debug("New entry: key l: " + key.length  + ", value: " + value.length);
```
spinscale: can you wrap all those kind of logger statements into `if (logger.isDebugEnabled())` - otherwise the string concatenation happens on every run (because the string is created, before the logger decides if this should be logged or not), stealing you some performance

spinscale: Elasticsearch has an alternative logger, which is not doing strin concatenation like `logger.log("{} is {}", key.length, value.length)", but I dont know if logstash has such...

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private boolean needAck(Message message) {
+        return message.getSequence() == message.getBatch().getBatchSize();
+    }
+
+    private void ack(ChannelHandlerContext ctx, Message message) {
+        writeAck(ctx, message.getBatch().getProtocol(), message.getSequence());
+    }
+
+    public void writeAck(ChannelHandlerContext ctx, byte protocol, int sequence) {
```
spinscale: private?

src/main/java/org/logstash/beats/CompressedBatchEncoder.java
```diff
@@ -0,0 +1,38 @@
...
+        Deflater deflater = new Deflater();
+        ByteBufOutputStream output = new ByteBufOutputStream(ctx.alloc().buffer());
+        DeflaterOutputStream outputDeflater = new DeflaterOutputStream(output, deflater);
+
+        byte[] chunk = new byte[payload.readableBytes()];
+        payload.readBytes(chunk);
+        outputDeflater.write(chunk);
+        outputDeflater.close();
+
+        ByteBuf content = ctx.alloc().buffer();
```
spinscale: you could allocate with a certain size?

src/main/java/org/logstash/beats/Runner.java
```diff
@@ -0,0 +1,46 @@
...
+
+        // Check for leaks.
+        // ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.PARANOID);
+
+        Server server = new Server(DEFAULT_PORT);
+
+        if(args.length > 0 && args[0].equals("ssl")) {
+            logger.debug("Using SSL");
+
+            String sslCertificate = "/Users/ph/es/certificates/certificate.crt";
```
spinscale: left over with static paths? Netty has a small helper class called `SelfSignedCertificate` that allows building certs on the fly for testing

src/main/java/org/logstash/beats/Server.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public Server(int p) {
+        port = p;
+        bossGroup = new NioEventLoopGroup();
+        workGroup = new NioEventLoopGroup();
+
+
+    }
+
+    public void enableSSL(SslSimpleBuilder builder) {
```
spinscale: should be part of the ctor (you cant change it later, right?)

src/main/java/org/logstash/beats/Server.java
```diff
@@ -0,0 +1,137 @@
...
+
+        Future<?> bossWait = bossGroup.shutdownGracefully(0, SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        Future<?> workWait = workGroup.shutdownGracefully(0, SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+
+        bossWait.await();
+        workWait.await();
+        logger.debug("Server stopped");
+    }
+
+    public void setMessageListener(IMessageListener listener) {
```
spinscale: move into ctor as well?

src/main/java/org/logstash/beats/Server.java
```diff
@@ -0,0 +1,137 @@
...
+        public BeatsInitializer(Server s) {
+            server = s;
+            beatsHandler = new BeatsHandler(server.messageListener);
+            idleExecutorGroup = new DefaultEventExecutorGroup(DEFAULT_IDLESTATEHANDLER_THREAD);
+        }
+
+        public void initChannel(SocketChannel socket) throws IOException, NoSuchAlgorithmException, CertificateException {
+            ChannelPipeline pipeline = socket.pipeline();
+
+            pipeline.addLast(LOGGER_HANDLER, loggingHandler);
```
spinscale: do you add the logging handler here and above as a child handler? why that?

src/main/java/org/logstash/beats/Server.java
```diff
@@ -0,0 +1,137 @@
...
+
+            if(server.isSslEnable()) {
+                SslHandler sslHandler = sslBuilder.build(socket.alloc());
+                pipeline.addLast(SSL_HANDLER, sslHandler);
+            }
+
+            // We have set a specific executor for the idle check, because the `beatsHandler` can be
+            // blocked on the queue, this the idleStateHandler manage the `KeepAlive` signal.
+            pipeline.addLast(idleExecutorGroup, KEEP_ALIVE_HANDLER, new IdleStateHandler(60*15, 5, 0));
```
spinscale: maybe set the arguments of idlestatehandler as variables before, so people know what they are for?

src/main/java/org/logstash/beats/Server.java
```diff
@@ -0,0 +1,137 @@
...
+        private final String KEEP_ALIVE_HANDLER = "keep-alive-handler";
+        private final String BEATS_PARSER = "beats-parser";
+        private final String BEATS_HANDLER = "beats-handler";
+        private final int DEFAULT_IDLESTATEHANDLER_THREAD = 4;
+        private final EventExecutorGroup idleExecutorGroup;
+        private final BeatsHandler beatsHandler;
+        private final LoggingHandler loggingHandler = new LoggingHandler();
+        private final Server server;
+
+        public BeatsInitializer(Server s) {
```
spinscale: you just need the messagelistener and a boolean as arguments

src/main/java/org/logstash/netty/PrivateKeyConverter.java
```diff
@@ -0,0 +1,76 @@
...
+import java.security.PrivateKey;
+import java.security.Security;
+import java.security.spec.InvalidKeySpecException;
+
+/*
+ * Take an Pem RSA Private key and convert it to a Pkcs8 private key that netty
+ * can understand.
+ *
+ */
+public class PrivateKeyConverter {
```
spinscale: IIRC a `PemReader` class in netty, maybe you can delete this one all together

spinscale:
* Gradle: Elasticsearch does not use the gradlew scripts. The distribution url contains a pretty old gradle version. This is needed in your case I guess, as it allows to not ship with gradle
* If you want to test further for leaks, there is also a Netty Leak Detector, that you can set to more restrictive settings (I think one is called paranoid, might be good for testing)
* There is a CircuitBreaker in old code? Guess this was impl specific and is not needed
* When a new `Message` is created, we need the `Map<String, String>` for JSON serialization. In the other case, we convert bytes to strings and back to bytes? I am not a hundred percent sure it that conversion is needed? Maybe I misread the code here though
* `BeatsHandler` happens on the netty event loop, same for the message listener. Just to be aware of, that these dont contain long running operations.
* Logging should be fixed
* Passphrase is in cleartext in memory (as it is in the config file so this is not an issue I guess)
* Do you need the `IMessageListener` interface with just a single impl? Could overwrite the needed methods in your tests anyway

ph:
@spinscale Thanks for taking the time for really valuable feedback!

build.gradle
ph: this should be cross platform, if its not present it will automatically fall back to the other implementation.

build.gradle
ph: @talevy This will be surely problematic someday, not sure if we have a good strategy around the jar hell.

src/main/java/org/logstash/beats/Message.java
ph: The problem here is the payload might not always be string in the context of the v2 protocol, filebeat. I think we rely in the code on jackson automatic understanding of json to create the right object, int, string or even other hash.

src/main/java/org/logstash/beats/BatchEncoder.java
ph: Yewps, left over, I will remove this and use the {} as the string interpolation strategy.

src/main/java/org/logstash/beats/Runner.java
ph: Well I needed to use the certificate outside of the java code for filebeat, I could have written it to disk. will change that.

src/main/java/org/logstash/beats/Server.java
ph: ctor?

ph:
> If you want to test further for leaks, there is also a Netty Leak Detector, that you can set to more restrictive settings (I think one is called paranoid, might be good for testing)
I've used it in the beginnign of the developpement will 

> There is a CircuitBreaker in old code? Guess this was impl specific and is not needed
Not needed anymore we have change strategy to send a ack 0 instead with the `IdleStateHandler`.

> When a new Message is created, we need the Map<String, String> for JSON serialization. In the other case, we convert bytes to strings and back to bytes? I am not a hundred percent sure it that conversion is needed? Maybe I misread the code here though
 
This is a particularity of the code between LSF and Filebeat, FB send a json object that may content more than just the <String, String>, it can contains anything from `int`, `string` or even hash.

> BeatsHandler happens on the netty event loop, same for the message listener. Just to be aware of, that these dont contain long running operations.

This might be the case since we don't have a non blocking queue for logstash, so things can possibly block under load. So my understanding was that no more events will be consumed when something block because all the threads in the event loop will be blocked. When the persistent queue this wont be a problem, might be worth to do a 1:1 to check how I can help that problem.


> Passphrase is in cleartext in memory (as it is in the config file so this is not an issue I guess)

This is a pretty good question, I could at least wrap it in a value/password so it doesn't leak in the log by error?

> Do you need the IMessageListener interface with just a single impl? Could overwrite the needed methods in your tests anyway
I will change it for that good suggestion.

src/main/java/org/logstash/beats/Server.java
spinscale: sorry, meant constructor

src/main/java/org/logstash/beats/BatchEncoder.java
ph: Thanks will verify that, I think the name of the plugin you are talking is `Buffer Monitor` I just installed it.

src/main/java/org/logstash/beats/Server.java
ph: Should only be one, left over when I've created the initializer class. good catch.

ph:
I've run this code for a few hours with `ResourceLeakDetector` set to `Paranoid` and I haven't seen anything in the log. I also monitored the memory using the VisualVm plugin for the buffer pools and I don't see any memory that was kept.

ph:
I am currently debugging the travis build, I am taking notes here to future reference..

- build jar, seems to be OK
- bundle install seems to be OK
- bundle exec rake vendor, doesnt work, it failled it cannot find the jars.

All theses commands are running locally, still trying to figure out what is the difference.

ph:
OK, Can reproduce locally.

rm -rf .mvn
rm -rf ~/.m2

ph:
Okay, after discussing with @talevy we agreed to change the strategy to make it work on travis.

ph:
From a discussion on zoom @spinscale and I agreed to do the following next steps to make sure the netty implementation is ready.

- Create a stress test that uses generatorbeats to create new connection and have multiple clients (100, 1000) and monitor the memory.
- implementat a way to show the direct memory usage in the logstash api that we could use to monitor the scenario.

spinscale:
Elasticsearch already monitors direct memory in the `JvmStats` class, see https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/monitor/jvm/JvmStats.java#L117

ph:
@spinscale Correct I will make a PR for core to have it by default in the api.

build.gradle
spinscale: cool, then we just need to release a new plugin version, if there is another sslpocalypse (hopefully much less often with boringssl)

lib/tasks/test.rake
```diff
@@ -0,0 +1,65 @@
+# encoding: utf-8
+OS_PLATFORM = RbConfig::CONFIG["host_os"]
+VENDOR_PATH = File.expand_path(File.join(File.dirname(__FILE__), "..", "..", "vendor"))
+
+if OS_PLATFORM == "linux"
+  FILEBEAT_URL = "https://beats-nightlies.s3.amazonaws.com/filebeat/filebeat-5.0.0-alpha4-SNAPSHOT-linux-x86_64.tar.gz"
```
spinscale: can you get the logstash version from somewhere and try to download the same filebeat version or is that unfeasible? otherwise you always need to update this one...

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -0,0 +1,104 @@
...
+    private void ack(ChannelHandlerContext ctx, Message message) {
+        writeAck(ctx, message.getBatch().getProtocol(), message.getSequence());
+    }
+
+    private void writeAck(ChannelHandlerContext ctx, byte protocol, int sequence) {
+        ByteBuf buffer = ctx.alloc().buffer(6);
+        buffer.writeByte(protocol);
+        buffer.writeByte('A');
+        buffer.writeInt(sequence);
+        ctx.writeAndFlush(buffer);
```
spinscale: does this bytebuffer potentially need to be released as well?

spinscale: another external test with many acks might help to spot this

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+
+                if(Protocol.isVersion2(currentVersion)) {
+                    logger.debug("Frame version 2 detected");
+                    batch.setProtocol(Protocol.VERSION_2);
+                } else {
+                    logger.debug("Frame version 1 detected");
+                    batch.setProtocol(Protocol.VERSION_1);
+                }
+
+                transition(States.READ_FRAME_TYPE, 1);
```
spinscale: minor cleanup, you could put the size of the types into the enum, so they are not magic numbers hanging around...

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+            case READ_WINDOW_SIZE: {
+                logger.debug("Running: READ_WINDOW_SIZE");
+
+                batch.setBatchSize((int) in.readUnsignedInt());
+
+                // This is unlikely to happen but I have no way to known when a frame is
+                // actually completely done other than checking the windows and the sequence number,
+                // If the FSM read a new window and I have still
+                // events buffered I should send the current batch down to the next handler.
+                if(!batch.isEmpty()) {
```
spinscale: implementation detail question here: Should this actual return an error to the client as he potentially has done sth wrong? Or are we good to go and live our lifes?

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+                break;
+            }
+            case READ_DATA_FIELDS: {
+                // Lumberjack version 1 protocol, which use the Key:Value format.
+                logger.debug("Running: READ_DATA_FIELDS");
+                sequence = (int) in.readUnsignedInt();
+                int fieldsCount = (int) in.readUnsignedInt();
+                int count = 0;
+
+                Map dataMap = new HashMap<String, String>();
```
spinscale: you can size the hashmap appropriately by putting `fieldsCount` as constructor argument

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+
+            case READ_COMPRESSED_FRAME: {
+                logger.debug("Running: READ_COMPRESSED_FRAME");
+
+
+                byte[] bytes = new byte[(int) requiredBytes];
+                in.readBytes(bytes);
+
+                InputStream inflater = new InflaterInputStream(new ByteArrayInputStream(bytes));
+                ByteArrayOutputStream decompressed = new ByteArrayOutputStream();
```
spinscale: netty has a `ByteBufOutputStream`, so you can write directly to a bytebuf.. and there is also a `ByteBufInputStream`, so you could try not to create a copy of the byte area in memory

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+
+
+                byte[] chunk = new byte[CHUNK_SIZE];
+                int length = 0;
+
+                while ((length = inflater.read(chunk)) > 0) {
+                    decompressed.write(chunk, 0, length);
+                }
+
+                inflater.close();
```
spinscale: if you are on java 7 in logstash, you could use the try-with-resources here, to ensure the resources get closed https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/org/logstash/beats/BeatsParser.java
spinscale: Another thing that comes to mind here and might be worth testing to prevent copying byte arrays around too much:

use a `InflaterOutputStream` and a `ByteBufOutputStream` and call `in.readBytes(OutputStream os, int length)`

src/main/java/org/logstash/beats/CompressedBatchEncoder.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class CompressedBatchEncoder extends BatchEncoder {
+    private final static Logger logger = LogManager.getLogger(BatchEncoder.class.getName());
+
+    @Override
+    protected ByteBuf getPayload(ChannelHandlerContext ctx, Batch batch) throws IOException {
+        ByteBuf payload = super.getPayload(ctx, batch);
+
+        Deflater deflater = new Deflater();
+        ByteBufOutputStream output = new ByteBufOutputStream(ctx.alloc().buffer());
```
spinscale: again try-with-resources here to ensure that every resource gets closed, no matter if exceptions are thrown

src/main/java/org/logstash/beats/BatchEncoder.java
```diff
@@ -0,0 +1,81 @@
...
+
+        try {
+            out.writeBytes(buffer);
+        } finally {
+            buffer.release();
+        }
+    }
+
+    protected ByteBuf getPayload(ChannelHandlerContext ctx, Batch batch) throws IOException {
+        ByteBuf payload = ctx.alloc().buffer();
```
spinscale: do you know the size you need to allocate here?

src/main/java/org/logstash/beats/CompressedBatchEncoder.java
```diff
@@ -0,0 +1,40 @@
...
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufOutputStream;
+import io.netty.channel.ChannelHandlerContext;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import java.io.IOException;
+import java.util.zip.Deflater;
+import java.util.zip.DeflaterOutputStream;
+
+public class CompressedBatchEncoder extends BatchEncoder {
```
spinscale: where is this one used? I seem to miss it.

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+                }
+
+                break;
+            }
+            case READ_JSON: {
+                logger.debug("Running: READ_JSON");
+
+                byte[] bytes = new byte[(int) requiredBytes];
+                in.readBytes(bytes);
+                Message message = new Message(sequence, (Map) JsonUtils.mapper.readValue(bytes, Object.class));
```
spinscale: what happens where there is invalid JSON in the bytes array?

src/main/java/org/logstash/netty/SslSimpleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+
+    public InputStream getSslCertificateFile() {
+        return sslCertificateFile;
+    }
+
+    public SslHandler build(ByteBufAllocator bufferAllocator) throws IOException, NoSuchAlgorithmException, CertificateException {
+        SslContextBuilder builder = SslContextBuilder.forServer(sslCertificateFile, sslKeyFile, passPhrase);
+
+
+        logger.debug("Ciphers: " + String.join(",", ciphers));
```
spinscale: other logger syntax to prevent always happening string concatenation

src/main/java/org/logstash/netty/SslSimpleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+        builder.ciphers(Arrays.asList(ciphers));
+
+        if(requireClientAuth()) {
+            logger.debug("Certificate Authorities: " + String.join(", ", certificateAuthorities));
```
spinscale: same here

src/main/java/org/logstash/netty/SslSimpleBuilder.java
```diff
@@ -0,0 +1,169 @@
...
+            builder.trustManager(loadCertificateCollection(certificateAuthorities));
+        }
+
+        SslContext context = builder.build();
+        SslHandler sslHandler = context.newHandler(bufferAllocator);
+
+        logger.debug("TLS: " +  String.join(",", protocols));
```
spinscale: you guessed it :-)

src/test/java/org/logstash/beats/BatchTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    public void TestGetDefaultProtocol() {
+        Batch batch = new Batch();
+        assertEquals(Protocol.VERSION_2, batch.getProtocol());
+    }
+
+    @Test
+    public void TestGetSetProtocol() {
+        Batch batch = new Batch();
```
spinscale: same test as above

src/main/java/org/logstash/beats/BeatsParser.java
ph: Oh crap forgot about that.

src/test/java/org/logstash/beats/BeatsParserTest.java
```diff
@@ -0,0 +1,103 @@
...
+        channel.writeInbound(o);
+
+        Batch decodedBatch = (Batch) channel.readInbound();
+
+        assertMessages(this.batch, decodedBatch);
+    }
+
+    private void assertMessages(Batch expected, Batch actual) {
+        assertNotNull(actual);
+        assertEquals(expected.getMessages().get(0).getSequence(), actual.getMessages().get(0).getSequence());
```
spinscale: how about `assertEquals(expected.getMessages(), actual.getMessages())`?

src/test/java/org/logstash/beats/BeatsParserTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        this.message1 = new Message(1, map1);
+        this.batch.addMessage(this.message1);
+
+        for(int i = 0; i < numberOfMessage; i++) {
+            Map map2 = new HashMap<String, String>();
+            map2.put("line", "Another world");
+            map2.put("from", "Little big Adventure");
+
+            this.message2 = new Message(i + 1, map2);
```
spinscale: why do you loop, but always overwrite message2? intended? if so, feels hard to read :-)

src/test/java/org/logstash/beats/BeatsParserTest.java
```diff
@@ -0,0 +1,103 @@
...
+            this.batch.addMessage(this.message2);
+        }
+    }
+
+    @Test
+    public void testEncodingDecodingJson() {
+        EmbeddedChannel channel = new EmbeddedChannel(new BatchEncoder(), new BeatsParser());
```
spinscale: you can create the channel as part of the setup and close it as part of the teardown?

src/test/java/org/logstash/beats/BeatsParserTest.java
```diff
@@ -0,0 +1,103 @@
...
+        Batch decodedBatch = channel.readInbound();
+        assertMessages(this.batch, decodedBatch);
+    }
+
+
+    @Test
+    public void  testCompressedEncodingDecodingFields() {
+        this.batch.setProtocol(Protocol.VERSION_1);
+
+        EmbeddedChannel channel = new EmbeddedChannel(new CompressedBatchEncoder(), new BeatsParser());
```
spinscale: these four lines repeat in all tests, refactor into own method?

src/main/java/org/logstash/beats/BeatsParser.java
ph: Thats a pretty good question, in the current implementation the following will happen:

- An exception will be raised
- The handler will catch it log it.
-  the whole frame will be discarded
- This is what is currently happening in the current ruby implementation.


So this might not be ideal, logstash doesn't have the concept of a DLQ. The protocol does ack on batch so If one of the event cannot be processed we cannot ack the batch.

src/main/java/org/logstash/beats/BeatsParser.java
ph: wonderful will make the changes!

src/main/java/org/logstash/beats/BeatsParser.java
ph: Agree

src/main/java/org/logstash/beats/BeatsParser.java
ph: This would require changes to the clients to have more bidirectional messages. So for the first implementation I think we are fine.

src/main/java/org/logstash/beats/BeatsHandler.java
ph: From the doc, I pretty sure I read that buffer will be release when the write occur?

spinscale: indeed, just looked it up

src/main/java/org/logstash/beats/BeatsParser.java
ph: @jordansissel What do you think  we should do in that case? Log the event and go on?

lib/tasks/test.rake
ph: Currently these is no generic name for it, I will recheck with the team if it changed.

src/main/java/org/logstash/beats/CompressedBatchEncoder.java
ph: Only used in the BeatsParserTest

src/main/java/org/logstash/netty/SslSimpleBuilder.java
ph: good catch

src/test/java/org/logstash/beats/BeatsParserTest.java
ph: I will change the expectation and the setup of the test to make it more clear.

src/test/java/org/logstash/beats/BeatsParserTest.java
ph: make sense doing it.

src/main/java/org/logstash/beats/BatchEncoder.java
ph: I don't really know upfront before serializing (json or fields), Any recommandation. Any idea how I could approximate that value?

src/test/java/org/logstash/beats/BeatsParserTest.java
ph: Yeah, well I change the expectation to do a more in details check between all the messages. It make more sense. In a future version we could randomize the data set.

ph:
I did another pass @spinscale !

src/main/java/org/logstash/beats/BatchEncoder.java
spinscale: indeed, not sure if a useful approximation exists, messages can be very different in size. ignore this comment

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -25,14 +29,21 @@
...
+        READ_HEADER(1),
+        READ_FRAME_TYPE(1),
+        READ_WINDOW_SIZE(4),
+        READ_JSON_HEADER(8),
+        READ_COMPRESSED_FRAME_HEADER(4),
+        READ_COMPRESSED_FRAME(-1), // -1 means the length to read is variable and defined in the frame itself.
+        READ_JSON(-1),
+        READ_DATA_FIELDS(-1);
+
+        private int value;
```
spinscale: name it `length`?

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -0,0 +1,230 @@
...
+
+        private States(int v) {
+            value = v;
+        }
+
+    }
+
+    private States currentState = States.READ_HEADER;
+    private long requiredBytes = 0;
```
spinscale: this could be an int? so there is no need for casting below

src/main/java/org/logstash/beats/CompressedBatchEncoder.java
spinscale: move it to the tests package then? or useful here as well?

spinscale:
left some minor comments, but I like this. hope to get this into more testing soon! :)

spinscale:
quick update: Be careful with jackson and dependencies.. I just checked out this issue in the rss plugin and I see a reference to afterburner already https://github.com/logstash-plugins/logstash-input-rss/issues/21

jordansissel:
Failing to use this plugin:

```
% bin/logstash -e 'input { beats { } }'
fetched an invalid config {:config=>"input { beats { } }output { stdout { codec => rubydebug } }", :reason=>"Couldn't find any input plugin named 'beats'. Are you sure this is correct? Trying to load the beats input plugin resulted in this error: no such file to load -- vendor/jar-dependencies/logstash-input-beats-3.1.0.beta1", :level=>:error}
```

Noting the 'no such file to load', I checked a few things and everything seems OK to me:

```
% grep logstash-input-beats Gemfile*
Gemfile:gem "logstash-input-beats", "3.1.0.beta1", :path => "vendor/local_gems/18ae8921/logstash-input-beats-3.1.0.beta1-java"
Gemfile.jruby-1.9.lock:  remote: vendor/local_gems/18ae8921/logstash-input-beats-3.1.0.beta1-java
Gemfile.jruby-1.9.lock:    logstash-input-beats (3.1.0.beta1-java)
Gemfile.jruby-1.9.lock:  logstash-input-beats (= 3.1.0.beta1)!
```

^^^ Correct version installed, it seems.

And the file it should have tried to load does exist:

```
vendor/local_gems/18ae8921/logstash-input-beats-3.1.0.beta1-java/vendor/jar-dependencies/logstash-input-beats-3.1.0.beta1.jar
```

Weird?

lib/logstash/inputs/beats.rb
```diff
@@ -2,21 +2,16 @@
...
-require "logstash/inputs/beats_support/circuit_breaker"
-require "logstash/inputs/beats_support/codec_callback_listener"
-require "logstash/inputs/beats_support/connection_handler"
-require "logstash/inputs/beats_support/event_transform_common"
-require "logstash/inputs/beats_support/decoded_event_transform"
-require "logstash/inputs/beats_support/raw_event_transform"
-require "logstash/inputs/beats_support/synchronous_queue_with_offer"
 require "logstash/util"
-require "thread_safe"
+require "vendor/jar-dependencies/logstash-input-beats-3.1.0.beta1.jar"
```
jordansissel: This causes Logstash to fail to load the plugin because the file is not found because it's the wrong path.

The .gemspec puts `vendor/jar-dependencies` into ruby's library path, so you can just do `require "logstash-input-beats-3.1.0.beta1.jar"` here. If I make this change, I can load this plugin into logstash.

jordansissel:
If you throw garbage at a plain text beats input, you get all kinds of fun exceptions:

To reproduce:

```
% bin/logstash --log.level debug -e 'input { beats { port => 12345 } }'
...
% dd if=/dev/urandom bs=4k count=1 | nc localhost 12345
```

Sample exception logged by Logstash:
```
Jul 11, 2016 3:05:14 PM io.netty.channel.AbstractChannelHandlerContext invokeExceptionCaught
WARNING: An exception was thrown by a user handler's exceptionCaught() method while handling the following exception:
io.netty.handler.codec.DecoderException: java.lang.IllegalArgumentException: Illegal initial capacity: -1982301064
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:442)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
        at io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:36)
        at io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:339)
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:339)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:393)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:742)
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:145)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.IllegalArgumentException: Illegal initial capacity: -1982301064
        at java.util.HashMap.<init>(HashMap.java:448)
        at java.util.HashMap.<init>(HashMap.java:467)
        at org.logstash.beats.BeatsParser.decode(BeatsParser.java:125)
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:411)
        ... 9 more
```

jordansissel:
Sending a MAX_INT as the json payload length causes fun errors>

Reproduction via ruby:

```
>> o = { "hello" => "world" }; TCPSocket.new("localhost", 12345).tap { |x| x.write(["1W", 1].pack("A*N")); x.write(["2J", 1, -1, o.to_json].pack("A*NNA*")); }.close
```

Result:

```
15:19:36.794 [nioEventLoopGroup-3-2] ERROR org.logstash.beats.BeatsHandler - Exception
io.netty.handler.codec.DecoderException: java.lang.NegativeArraySizeException
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:442) ~[netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248) ~[netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:36) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:339) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:339) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:393) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:742) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:145) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at java.lang.Thread.run(Thread.java:745) [?:1.8.0_72]
Caused by: java.lang.NegativeArraySizeException
        at org.logstash.beats.BeatsParser.decode(BeatsParser.java:193) ~[logstash-input-beats-3.1.0.beta1.jar:?]
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:411) ~[netty-all-4.1.1.Final.jar:4.1.1.Final]
        ... 9 more
```

jordansissel:
Hilarious things happen when Logstash runs out of file descriptors. I triggered this scenario by simply opening lots of connections.

```
>> 10000.times.collect { TCPSocket.new("localhost", 12345) }.map(&:close)
```

.gitignore
```diff
@@ -21,3 +21,12 @@ logstash-forwarder
 *.json
 vendor/
 .bundle/
+build/
+.idea
+.gradle
+null
```
jordansissel: null? hah.

jordansissel:
Very poorly-scienced performance comparison on my 1-core VM

```
% (cd logstash-5.0.0-alpha4-with-new-beats; bin/logstash-plugin list --verbose | grep beats; bin/logstash -e 'input { beats { port => 12345 } } output { stdout { codec => dots } }'  | pv -War > /dev/null)
logstash-input-beats (3.1.0.beta1)
...
[13.5KiB/s] [9.37KiB/s]
```

Peak was around 13.5Keps (this PR)

```
% (cd logstash-5.0.0-alpha4; bin/logstash-plugin list --verbose | grep beats; bin/logstash -e 'input { beats { port => 12345 } } output { stdout { codec => dots } }'  | pv -War > /dev/null)
logstash-input-beats (3.0.3)
...
[11.2KiB/s] [4.55KiB/s]
```

Release (5.0.0-alpha4) beats input peaks at 11.2Keps.

<details>
<summary>filebeat.yml</summary>
```
% egrep -v '^\s*#|^\s*$' filebeat.yml
filebeat:
  prospectors:
    -
      paths:
        - /var/log/*.log
      input_type: stdin
output:
  logstash:
    hosts: ["localhost:12345"]
shipper:
logging:
  files:
    rotateeverybytes: 10485760 # = 10MB
```
</summary>

jordansissel:
Same test as above with a 2-core VM (same physical machine)

* this pr: ~21Keps
* version 3.0.3: ~18Keps

I need to test against older versions also.

jordansissel:
I also get this sometimes when shutting down:

```
java.util.concurrent.RejectedExecutionException: event executor terminated
```

ph:
answering @jordansissel's comment here.


`java.util.concurrent.RejectedExecutionException: event executor terminated` I think this is due to the idle check execution pool, I was able to get this error in tests.


Concerning the test what did you use to generate the events, Filebeat from stdin?

jordansissel:
@ph yeah I used filebeat from stdin

ph:
@jordansissel My test were done with logstash using all available core, the null output and generator beats which just send a infinite number of events. The stats were gathered at the sending end using libbeat http server. The metric was the number of events ack per seconds from a 30s window. IIRC.

ph:
I am investigating the leak that @talevy has found when running the code locally, with the `LeakDetector` set to Advanced.

ph:
@jordansissel still using stdout represent more a real world case. :)

ph:
@jordansissel Concerning the Number of FD and when we send garbage, would you be OK to do that in another PR, this one is starting to be pretty big :)

ph:
@jordansissel did you use pipelining on the beat side under the logstash output configuration?
All the test we did was using 5 for the value, I don't think this option is on by default.
```
output.logstash:
  compression_level: 3
  hosts:
    - localhost:5044
  pipelining: 5
```

talevy:
LGTM.

I think this should be merged into `master` as is and we can open up new issues around @jordansissel to keep track of those separately: https://github.com/logstash-plugins/logstash-input-beats/pull/93#issuecomment-231903161

suyograo:
+1, this PR is getting huge as is.

@ph needs a rebase though

ph:
I've created issues from @jordansissels comments:

#100, #99, #98

PR #94, opened 2016-06-21
Fix description and edit SSL options

dedemorton:
@suyograo Alvin reminded me that we needed to update the description of the beats input plugin. Here are my changes/edits.


suyograo:
LGTM

PR #105, opened 2016-07-20
SSL Handshake Timeout

### LINKED ISSUE #101: Make ssl/socket acceptance have timeouts #10
copied from https://github.com/elastic/ruby-lumberjack/issues/10

We shouldn't allow a misbehaving client to hinder acceptance of new ssl connections. Reported in elastic/logstash-forwarder#429 is an issue (I think?) that relates to a bad/stalled client disappearing in the middle of of a handshake or is unable to complete it successfully - can cause the whole acceptor thread to stall, possibly forever.

ph:
SSL Handshake from client can sometime be stuck forever this PR expose
the internal netty ssl handler timeout and allow users to configure it.

By default Netty will close any stuck handshake that goes over 10000
milliseconds.

Fixes #101

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -86,6 +86,9 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   # This option needs to be used with `ssl_certificate_authorities` and a defined list of CAs.
   config :ssl_verify_mode, :validate => ["none", "peer", "force_peer"], :default => "none"
 
+  # Time in milliseconds for an incomplete ssl handshake to timeout
+  config :ssl_handshake_timeout, :validate => :number, :default => 10000
+
   # The number of seconds before we raise a timeout. 
   # This option is useful to control how much time to wait if something is blocking the pipeline.
   config :congestion_threshold, :validate => :number, :default => 5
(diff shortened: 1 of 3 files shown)


lib/logstash/inputs/beats.rb
```diff
@@ -86,6 +86,9 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   # This option needs to be used with `ssl_certificate_authorities` and a defined list of CAs.
   config :ssl_verify_mode, :validate => ["none", "peer", "force_peer"], :default => "none"
 
+  # Time in milliseconds for an incomplete ssl handshake to timeout
+  config :ssl_handshake_timeout, :validate => :number, :default => 10000
```
talevy: is there a test for when you exceed the timeout?

ph: I am actually exposing a SSLHandler settings down to the consumer I though adding a test for it was unnecessary. I could be wrong WDYT?

ph: Actually, Netty's SSLHandler has already a sane default of 10000ms.

ph:
I've added test for it and found a bug when the client timeout is hit before the ssl handshake timeout, I am looking at the option sets on the `SSLEngine`.

The test case use the server and a netty client, I dont recall seen this when testing manually with the `TCPConnect` in ruby.

ph:
@talevy This is pretty complex scenario to test, even Netty has no tests for it that I can find :(

The way to test this is to setup a server with SSL and a client with SSL also and make the handshake fail due a timeout.. which is I haven't found an easy way to do.

I see this problem complementary to #108 which will close idle connection.

ph:
@talevy Update on this?

talevy:
LGTM

Issue #100, opened 2016-07-19
Sending garbage data will raise weird exception


ph:
If you throw garbage at a plain text beats input, you get all kinds of fun exceptions:

To reproduce:

% bin/logstash --log.level debug -e 'input { beats { port => 12345 } }'
...
% dd if=/dev/urandom bs=4k count=1 | nc localhost 12345

full output
```
Jul 11, 2016 3:05:14 PM io.netty.channel.AbstractChannelHandlerContext invokeExceptionCaught
WARNING: An exception was thrown by a user handler's exceptionCaught() method while handling the following exception:
io.netty.handler.codec.DecoderException: java.lang.IllegalArgumentException: Illegal initial capacity: -1982301064
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:442)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
        at io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:36)
        at io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:339)
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:339)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:393)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:742)
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:145)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.IllegalArgumentException: Illegal initial capacity: -1982301064
        at java.util.HashMap.<init>(HashMap.java:448)
        at java.util.HashMap.<init>(HashMap.java:467)
        at org.logstash.beats.BeatsParser.decode(BeatsParser.java:125)
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:411)
        ... 9 more
```

ph:
[...]

Issue #99, opened 2016-07-19
Sending a MAX_INT as the json payload length cause the parser to throw an exception


ph:
Sending a MAX_INT as the json payload length causes fun errors

Reproduction:

```
>> o = { "hello" => "world" }; TCPSocket.new("localhost", 12345).tap { |x| x.write(["1W", 1].pack("A*N")); x.write(["2J", 1, -1, o.to_json].pack("A*NNA*")); }.close
```

full output
```
15:19:36.794 [nioEventLoopGroup-3-2] ERROR org.logstash.beats.BeatsHandler - Exception
io.netty.handler.codec.DecoderException: java.lang.NegativeArraySizeException
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:442) ~[netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248) ~[netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:36) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:339) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:339) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:393) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:742) [netty-all-4.1.1.Final.jar:4.1.1.Final]
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:145) [netty-all-4.1.1.Final.jar:4.1.1.Final]
[...]

PR #106, opened 2016-07-20
Gracefully recover from sending malformed dataframe

ph:
This change make sure the state machien correctly recover from sending
garbage data over the write. The current strategy is to check to make
sure the frametype are correctly send if we received an unknown frame
type we will assume that the frame is poisonous, we will raise a custom
exception `InvalidFrameProtocol` with the associated bytes and will
terminate the connection since we cannot do anything with it.

The BeatsParser will now raise an exception when we receive an invalid size for the json payload (v2) or the fieldcounts (v1)

Fixes: #100 #99

### DIFF
--- a/src/main/java/org/logstash/beats/BeatsParser.java
+++ b/src/main/java/org/logstash/beats/BeatsParser.java
@@ -4,23 +4,17 @@ package org.logstash.beats;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufInputStream;
 import io.netty.buffer.ByteBufOutputStream;
-import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.ByteToMessageDecoder;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.zip.Inflater;
-import java.util.zip.InflaterInputStream;
 import java.util.zip.InflaterOutputStream;
 
 
@@ -96,6 +90,9 @@ public class BeatsParser extends ByteToMessageDecoder {
                     case Protocol.CODE_FRAME: {
                         transition(States.READ_DATA_FIELDS);
                         break;
+                    }
+                    default: {
+                        throw new InvalidFrameProtocol(frameType);
                     }
                 }
                 break;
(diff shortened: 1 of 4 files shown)


src/test/java/org/logstash/beats/BeatsParserTest.java
```diff
@@ -50,14 +53,25 @@ public void testEncodingDecodingFields() {
...
     @Test
-    public void  testCompressedEncodingDecodingFields() {
+    public void testCompressedEncodingDecodingFields() {
         this.batch.setProtocol(Protocol.VERSION_1);
         Batch decodedBatch = decodeCompressedBatch();
         assertMessages(this.batch, decodedBatch);
     }
 
+    @Test
+    public void testShouldNotCrashOnGarbageData() {
```
talevy: add a test for when `InvalidFrameProtocol` is thrown? something like verifying that the `exceptionCaught` was kicking in?

ph: Yes, I will do that, I am actually leaning toward removing the exception caught in the BeatsParser and add a new handler to only deal with that and would make code easier to test?

talevy: I think it is fine to keep it together since it is so tightly coupled and it is just one exception.

ph: I've changed the expectation and the exception is now propagated down stream in the BeatsHandler, which should have been in the first place.

ph: Just to clarify that, the exception **need** to be propagated down to the `BeatsHandler`, because the handler will notify the ruby listener so the connections can be correctly removed from the connection list and the buffered data flushed down the pipeline.

ph:
@talevy I have updated this to include a fix for #99, It was better to do it in this one because the errors are now propagated downstream.

Also, I've created specific methods for creating invalid data frame which arent a part of the `BatchEncoder`, I believe they don't belong in that class since this class could be used to create a real client.

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -232,4 +237,16 @@ private void batchComplete() {
...
         batch = new Batch();
     }
+
+    public class InvalidFrameProtocol extends Exception {
+        public InvalidFrameProtocol(byte frameType) {
+            super("Invalid Frame Protocol, Received: " + frameType);
+        }
+    }
+
+    public class InvalidPayloadSize extends Exception {
```
talevy: best to keep the `Exception` suffix

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -232,4 +237,16 @@ private void batchComplete() {
         sequence = 0;
         batch = new Batch();
     }
+
+    public class InvalidFrameProtocol extends Exception {
```
talevy: best to keep the `Exception` suffix

talevy:
Do we need a special exception type for each one of those frame scenarios? maybe consolidate them all into InvalidParseException? or something like that, and have the proper message passed in

src/test/java/org/logstash/beats/BeatsParserTest.java
```diff
@@ -50,22 +61,139 @@ public void testEncodingDecodingFields() {
...
+        payload.writeInt(key.length);
+        payload.writeBytes(key);
+        payload.writeInt(value.length);
+        payload.writeBytes(value);
+
+        sendPayloadToParser(payload);
+    }
+
+    private void sendInvalidJSonPayload(int size) throws JsonProcessingException {
+        thrown.expectCause(isA(BeatsParser.InvalidPayloadSize.class));
```
talevy: do we want to also test the message contents?

talevy:
@ph left some small comments, but looks good

ph:
You are right I will consolidate the Exception into one, since we dont have a lot of them this makes sense.

ph:
changes done, can you take another look?

talevy:
LGTM

ph:
rebased

Issue #74, opened 2016-04-11
TCP connections not being closed causes FD leak


robin13:
As there is not (currently) a keepalive signal from beats or a timeout in beats, connections remain indefinitely open leading to a filedescriptor leak - over time a logstash agent running with the logstash-input-beats may have thousands of open TCP connections, eventually running against the system limit for filedescriptors.

kurt5:
+1

jordansissel:
@robin13 ahh, if the connection is opened by beats (client) and then abandoned (due to network failure or other disappearing-act) the TCP connection never closes? 

+1 on adding a heartbeat frame to the beats protocol.

robin13:
@jordansissel Yes - that's it.

ph:
This is fixed in the java rewrite, the settings is very high for now 750 seconds, but I will reduce that to 30 and expose it in the plugin.

PR #108, opened 2016-07-21
Close idle client connection

ph:
The server will close the client connection after a certain amount of
activity, the default is 15 seconds but users can configure it at the
plugin level by setting the `client_inactivity_timeout` option to the desired
value.

Doing this will ensure that unterminated connection wont make logstash
run out of FD.

Fixes: #74

### DIFF
--- a/src/main/java/org/logstash/beats/Server.java
+++ b/src/main/java/org/logstash/beats/Server.java
@@ -10,13 +10,12 @@ import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.handler.logging.LoggingHandler;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.timeout.IdleStateHandler;
-import io.netty.util.concurrent.EventExecutorGroup;
 import io.netty.util.concurrent.DefaultEventExecutorGroup;
+import io.netty.util.concurrent.EventExecutorGroup;
 import io.netty.util.concurrent.Future;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.logstash.netty.SslSimpleBuilder;
-import javax.net.ssl.SSLException;
 import java.io.IOException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
@@ -29,10 +28,13 @@ public class Server {
     static final long SHUTDOWN_TIMEOUT_SECONDS = 10;
 
     private int port;
+
     private NioEventLoopGroup bossGroup;
     private NioEventLoopGroup workGroup;
     private IMessageListener messageListener = new MessageListener();
     private SslSimpleBuilder sslBuilder;
+
+    private int clientInactivityTimeoutSeconds = 15;
 
     public Server(int p) {
         port = p;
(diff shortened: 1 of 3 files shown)


src/main/java/org/logstash/beats/Server.java
```diff
@@ -29,11 +28,14 @@
     static final long SHUTDOWN_TIMEOUT_SECONDS = 10;
 
     private int port;
+
     private NioEventLoopGroup bossGroup;
     private NioEventLoopGroup workGroup;
     private IMessageListener messageListener = new MessageListener();
     private SslSimpleBuilder sslBuilder;
 
+    private int clientInactivityTimeoutSeconds = 15;
```
talevy: best to set the value in the constructor

src/main/java/org/logstash/beats/Server.java
```diff
@@ -29,11 +28,14 @@
     static final long SHUTDOWN_TIMEOUT_SECONDS = 10;
 
     private int port;
```
talevy: not needed for this PR, but maybe all of these can be set in the constructor and made `final`?

src/main/java/org/logstash/beats/Server.java
```diff
@@ -69,6 +71,10 @@ public Server listen() throws InterruptedException {
         return this;
     }
 
+    public void setClientInactivityTimeout(int time) {
```
talevy: maybe just add this to the constructor and remove the setter?

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+            Long started = System.currentTimeMillis() / 1000L;
+
+            ChannelFuture f = b.connect("localhost", randomPort).sync();
+            f.channel().closeFuture().sync();
+
+            Long ended = System.currentTimeMillis() / 1000L;
+
+            int diff = (int) (ended - started);
+            assertThat(diff, is(greaterThanOrEqualTo(inactivityTime)));
```
talevy: should we have stricter equality around this? maybe

```
import static org.hamcrest.number.IsCloseTo.closeTo
...
assertThat(diff, is(closeTo(inactivityTime, 0.1)))
```

ph: +1, I wasn't aware of closeTo, I though about using between()

ph:
updated with the review comment

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -59,8 +61,9 @@ public void initChannel(SocketChannel ch) throws Exception {
 
             Long ended = System.currentTimeMillis() / 1000L;
 
-            int diff = (int) (ended - started);
-            assertThat(diff, is(greaterThanOrEqualTo(inactivityTime)));
+            double diff = ended - started;
+            assertThat(diff, is(closeTo(inactivityTime, 0)));
```
talevy: haha, `0`? is that error margin not the same as equivalence?

talevy:
great! looks good. just had a small question about the `closeTo` with `error=0`

src/test/java/org/logstash/beats/ServerTest.java
ph: bleh, I've commited the test to make it fails  will revert that.

ph:
@talevy revert  to 0.1 sorry for the hassle.

talevy:
LGTM

Issue #111, opened 2016-07-28
IdleStateHandler isn't `@Shareable` and make multi connection crash.


ph:
The IdleStateHandler isn't `@Shareable` and this mean that the same instance cannot be used by multiple channel .

In clear, this is really a big problem since the server cannot accept multiples distinct connections, it will raise an exception. 

This is a regression merged by #101 and we don't have any integration test for multiples clients.
I will fix the issue and create an integration test for it.

PR #112, opened 2016-07-28
IdleStateHandler is not `@Shareable` between channel.

ph:
This PR fix an issue introduced by #105 that was rejection new
connection on the server because the `IdleStateHandler` instance cannot
be shared between channels.

This also introduce a test to test concurrent connection on the server,
we never had a similar test before.

Gradle test will be also run explicitely on travis.

Fixes: #111


ph:
@talevy Its a flaky test, trying to get a bit more details from travis run.

ph:
fixed @talevy

talevy:
LGTM (unfortunate about the settings issue)

PR #113, opened 2016-07-28
Update protocol spec to align with implementation

CVTJNII:
The following pull updates the protocol spec with the following changes:
- Update version to 2 as the JSON frame is a v2 improvement per https://discuss.elastic.co/t/logstash-beats-input-lumberjack-protocol-v1-support-lifetime/56334
- Note that the version byte is ASCII per https://github.com/logstash-plugins/logstash-input-beats/blob/a2ad0ea45c5edf69828518d4fc82a22b2817e403/lib/lumberjack/beats/server.rb#L219-L220
- Note that the window frame size needs to be set before each frame of messages, found by troubleshooting reset_next_ack() errors while implementing the spec.

Not sure if the CLA is required for simple documentation fixes, so please let me know if it is.  Thanks.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -9,7 +9,7 @@ The needs that lead to this protocol are:
 
 ## Implementation Considerations
 
-# Lumberjack Protocol v1
+# Lumberjack Protocol v2
 
 ## Behavior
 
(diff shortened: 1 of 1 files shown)


untergeek:
CLA is necessary for even documentation changes, yes.  Please sign the CLA.

CVTJNII:
Okay, I'll work on getting it signed.  I need corporate approval so it will probably take a little bit.  Thanks.

ph:
@CVTJNII Changes LGTM, so I will gladly merge it when the CLA is signed.

PR #114, opened 2016-08-03
Add configuration option to disable codec tagging

### LINKED ISSUE #58: beats_input_codec_json_applied being added to tags
I have json strings being written to a log file, one per line and I'm using filebeats to ship off to logstash (doing this instead of the logstash input plugin because that seemed to miss lines randomly and filebeats doesn't)

Here is the basic config for the input in logstash

```
input {
  beats {
    codec => "json"
    host => "127.0.0.1"
    port => 5044
  }
}
```

My json object has a "tags" property that we use already and it seems that the beats input plugin adds `beats_input_codec_json_applied` to the tags property. I've had to right a filter to remove that tag (which also slows down logstash unnecessarily)

Is there any way it can avoid adding that tag to the object?

### LINKED ISSUE #96: Document "beats_input_codec_#{codec_name}_applied" tag
When using the logstash-input-beats, it will add a tag: "beats_input_codec_#{codec_name}_applied"
https://github.com/logstash-plugins/logstash-input-beats/blob/master/lib/logstash/inputs/beats/decoded_event_transform.rb#L14

This behavior is not documented, and current docs say the tag has no default value.  https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html#plugins-inputs-beats-tags)

sigmavirus24:

This provides the "include_codec_tag" configuration option to allow
users to disable the application of the beats_input_codec_*_applied tag.

Closes #58
Closes #96

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,6 +10,7 @@ Contributors:
 * Richard Pijnenburg (electrical)
 * Suyog Rao (suyograo)
 * Guy Boertje (guyboertje)
+* Ian Cordasco (sigmavirus24)
 
 Note: If you've sent us patches, bug reports, or otherwise contributed to
 Logstash, and you aren't on the list above and want to be, please let us know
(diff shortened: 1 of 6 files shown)


sigmavirus24:
Just signed the ICLA. Didn't realize that was necessary. =D

ph:
LGTM

ph:
@sigmavirus24 It will roll out in the next beta version

sigmavirus24:
Awesome @ph!

PR #119, opened 2016-08-24
NPE, Error in the log and make it work with Logstash 2.4

ph:
- Fix a NullPointer Exception https://github.com/elastic/logstash/issues/5756
 - Log4j ERROR will now be propagated upstream, IE: InvalidCertificate OR InvalidFrameType.
 - Relax constraints on multiline to make it work under 2.4
 - use the default set of ciphers


suyograo:
LGTM

Issue #55, opened 2016-03-08
type field not being used


ppf2:
Tested using LS 2.2.2's beat input.

```
input {
  beats {
    port => 5044
	type => "beats"
  }
}
```

The type field specified is not being honored.

https://www.elastic.co/guide/en/logstash/2.2/plugins-inputs-beats.html#plugins-inputs-beats-type

When sending events from the filebeat, it retains the default "log" value as the type value even though type is set to "beats" in the LS beat input.

To get this to work, I ended up setting document_type at the beats level, and just let it pass all the way through to the LS pipeline.

```
      # Type to be published in the 'type' field. For Elasticsearch output,
      # the type defines the document type these entries should be stored
      # in. Default: log
      document_type: my_beats_type
```

Not sure if this is the intended behavior (LS beats input ignores its type parameter), if so, let's update our documentation accordingly.

andrewkroh:
I think that is the intended behavior.

The documentation for [type](https://www.elastic.co/guide/en/logstash/2.2/plugins-inputs-beats.html#plugins-inputs-beats-type) states:

> If you try to set a type on an event that already has one (for example when you send an event from a shipper to an indexer) then a new input will not override the existing type. A type set at the shipper stays with that event for its life even when sent to another Logstash server.

ppf2:
Somehow I missed that paragraph :)  Yah, will be helpful to clarify in the doc that the type parameter for the beats input will never be used since it is always set in the beat (and to use document_type in the beats configuration instead).

andrewkroh:
There have been numerous users on Discuss that try to use `type => "xyz"` in their beat input config, and then they write conditional logic based on the `type` field that does not work as expected. 

They always miss the note in the docs or never saw it. Since `type` is a required field that is enforced by libbeat, can we just remove the `type` option from the beats input?

These are few examples where this has occurred:

https://discuss.elastic.co/t/winlogbeat-event-tagging-and-overriding-type-in-logstash/42539
https://discuss.elastic.co/t/logstash-lost-eventlog-from-winlogbeat/55360
https://discuss.elastic.co/t/topbeat-with-redis/34157
https://discuss.elastic.co/t/how-to-tag-log-files-in-filebeat-for-logstash-ingestion/44713/5
https://discuss.elastic.co/t/winlogbeat-not-outputting-to-logstash/57500/6

iceiilin:
[...]

PR #121, opened 2016-08-25
Add Beats config example and note about type setting

dedemorton:
- Added Beats config info (requested by the Beats team).
- Added a note to explain that the `type` setting is ignored. Here's the issue that prompted this change: https://github.com/logstash-plugins/logstash-input-beats/issues/55


andrewkroh:
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 3.1.1
+   - [Doc changes]
```
suyograo: Maybe say briefly what changed?

Issue #123, opened 2016-08-25
Good certificates/private key will be declared as invalid


ph:
The integration test should have pick this up, I believe the latest changes from #119 broke the plugin. There seems to be a problem with the handling of private key and the converter, it is used to convert from PKC7 to PKCS8.

```
Aug 25, 2016 9:40:00 AM io.netty.channel.ChannelInitializer exceptionCaught
WARNING: Failed to initialize a channel. Closing: [id: 0x3759a880, L:/127.0.0.1:5044 - R:/127.0.0.1:63831]
java.lang.IllegalArgumentException: Input stream not contain valid certificates.
	at io.netty.handler.ssl.SslContextBuilder.keyManager(SslContextBuilder.java:272)
	at io.netty.handler.ssl.SslContextBuilder.forServer(SslContextBuilder.java:99)
	at org.logstash.netty.SslSimpleBuilder.build(SslSimpleBuilder.java:107)
	at org.logstash.beats.Server$BeatsInitializer.initChannel(Server.java:132)
	at org.logstash.beats.Server$BeatsInitializer.initChannel(Server.java:99)
	at io.netty.channel.ChannelInitializer.channelRegistered(ChannelInitializer.java:68)

ph:
This error was detected when running the `Runner` class inside the java project, 2.4 doesn't expose that much information.

tsg:
Still failing for me with plugin version 3.1.1, not sure if I'm doing anything else wrong. I don't get any errors in the Logstash logs, but I uploaded the full logs [here](https://gist.github.com/tsg/5e8d1e9db0fe84bd393d165457f0d885) in case I'm missing something.

ph:
Fixed in 3.1.2. Make sure the private keys are in PKCS8.

PR #124, opened 2016-08-25
Fix a few issues with the private key handling and JDK7 compatibility.

ph:
This PR introduce a few changes:

- It removes the `PrivateKeyConverter`, in retrospect this class was only used to make the private key generated by ruby work on the in the integration suite. Because Netty only accepts PKCS8 as the format of the key, in the real world this is not a problem any newer version of openssl will output PKCS8.

- It make sure the code can target JDK 1.7, Removes String.join and lambdas

- Remove bouncycastle

Fixes: https://github.com/logstash-plugins/logstash-input-beats/issues/123


ph:
I will rebase.

talevy:
LGTM

ph:
@tsg This should fix the problem you have discovered and will also reduce any future problem with ssl since we will not temper too much with the files and will just let netty handle the problems.

PR #125, opened 2016-08-26
propagate handshake errors  and fix logging

ph:
- Move back to log4j1
- Make sure we propagate the handshake errors


talevy:
LGTM

PR #126, opened 2016-08-26
Make sure we use a relative path for the version so the doc gen can read

ph:
We need to use relative path to allow the doc generator to read the gemspec.


suyograo:
LGTM

PR #130, opened 2016-09-15
Use correct log4j logger call for 2.4 compat

suyograo:


lib/logstash/inputs/beats.rb
```diff
@@ -141,8 +141,8 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   def register
     # Logstash 2.4
-    if defined?(LogStash::Logging) && LogStash::Logging.respond_to?(:setup_log4j)
-      LogStash::Logging.setup_log4j(@logger)
+    if defined?(LogStash::Logging) && LogStash::Logger.respond_to?(:setup_log4j)
```
talevy:
```ruby
defined?(LogStash::Logger)
```

suyograo: crap, forgot that. good eye

talevy:
LGTM

Issue #127, opened 2016-08-29
Standardize SSL Options


ph:
We are trying as much as possible to standardize all our SSL options accross all of our products.

### Server Settings
As a server (X-Pack, Kibana, Logstash):

| Description       | Setting       | Default | Valid values |
| ----------------- | ------------- | ------- | ------------ |
| enabled setting   | `ssl.enabled` | `false` | `true`, `false` |
| verification mode | `ssl.verification_mode` | `full` | `none`, `certificate`, `full` |
| supported protocols | `ssl.supported_protocols` | `TLSv1.2`, `TLSv1.1`, `TLSv1` | `TLSv1`, `TLSv1.1`, `TLSv1.2`, `SSLv3` |
| client authentication | `ssl.client_authentication` | `none` | `none`, `optional`, `required` |
| cipher suites | `ssl.cipher_suites` | see list below | language specific |
| certificate authorities | `ssl.certificate_authorities` | language/system trusted CAs | array of paths to PEM encoded certificates |
| key | `ssl.key` | | path to PEM encoded key file |
| key password/passphrase | `ssl.key_passphrase` | | |
| certificate | `ssl.certificate` | | path to file containing server certificate and optionally includes cert chain |

Note: It should be ok to have an inconsistency in the supported values for `ssl.verification_mode` as Go only supports `none` and `full`.

### Supported Cipher Suites
| Product       | Cipher Suites |
| ----------- | ------------- |
| Logstash      | Update to latest [mozilla intermediate ciphers](https://wiki.mozilla.org/Security/Server_Side_TLS) |

ph:
Concerning the ciphers with the java rewrite I can confirm that the DEFAULTS ciphers are the one described in the Mozilla list.

nick-george:
If you're looking at modifying any of the SSL code, fixing issue #196 while you're at it would be much appreciated!

PR #134, opened 2016-09-16
make the integration test work in latest filebeat master branch

ph:
Filebeat changed his options for SSLs and made all the integration test
use plain text.

The changes were the following

- tls -> ssl
- certificate_key => key


ph:
We still need to apply the new name to this plugin https://github.com/logstash-plugins/logstash-input-beats/issues/127

ph:
superseeded by https://github.com/logstash-plugins/logstash-input-beats/pull/139

Issue #138, opened 2016-09-21
The plugin raise a TypeError when you are using a `passphrase`


ph:
The passphrase uses the `Password` class to hide the sensitive information to retrive the actual password the code need to call `#value` on the instance. 

The code doesn't do that making the java layer throw a `TypeError` since it was expecting a String.

ph:
fixed by #139

PR #139, opened 2016-09-21
Fix an issue when specifying a passphrase that was raising a TypeError

ph:
The passphrase uses the `Password` class to encapsulate the sensitive
information and make sure we dont write it to the log by accident, to
get the original value the plugin must call `#value` on the object.

The Java layer was expecting a string but was receive a Password
instance and was throwing an exception.

This commit make sure we use password and add an integration test for
it.

Fixes: #138

### DIFF
--- a/spec/integration/filebeat_spec.rb
+++ b/spec/integration/filebeat_spec.rb
@@ -33,12 +33,10 @@ describe "Filebeat", :integration => true do
 
   let(:filebeat_exec) { FILEBEAT_BINARY }
 
-  let_empty_tmp_file(:registry_file)
   let(:filebeat_config) do
     {
       "filebeat" => {
         "prospectors" => [{ "paths" => [log_file],  "input_type" => "log" }],
-        "registry_file" => registry_file,
         "scan_frequency" => "1s",
         "idle_timeout" => "1s"
       },
@@ -92,7 +90,7 @@ describe "Filebeat", :integration => true do
       end
 
       let(:input_config) do
-        super.merge({ 
+        super.merge({
           "ssl" => true,
           "ssl_certificate" => certificate_file,
           "ssl_key" => certificate_key_file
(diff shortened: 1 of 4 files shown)


spec/integration/filebeat_spec.rb
```diff
@@ -33,13 +33,12 @@
...
   let(:filebeat_exec) { FILEBEAT_BINARY }
 
-  let_empty_tmp_file(:registry_file)
   let(:filebeat_config) do
     {
       "filebeat" => {
         "prospectors" => [{ "paths" => [log_file],  "input_type" => "log" }],
-        "registry_file" => registry_file,
-        "scan_frequency" => "1s"
+        "scan_frequency" => "1s",
```
andrewvc: Why were properties in this file changed for this PR?

spec/integration/filebeat_spec.rb
```diff
@@ -106,8 +105,38 @@
 
       context "self signed certificate" do
         include_examples "send events"
+
+
+        # Refactor this to use Flores's PKI instead of openssl command line
```
andrewvc: Nice!

spec/integration/filebeat_spec.rb
```diff
@@ -134,10 +163,10 @@
             "output" => {
               "logstash" => {
                 "hosts" => ["#{host}:#{port}"],
-                "tls" => { 
+                "ssl" => {
```
andrewvc: Why the change from tls to ssl?

ph:
@andrewvc the properties needed to be changed to keep it in sync with what changed from filebeat.

spec/integration/filebeat_spec.rb
ph: `registry_file` doesnt work anymore with FB 5

spec/integration/filebeat_spec.rb
suyograo: Also this is just in the spec

andrewvc:
LGTM

Issue #141, opened 2016-09-22
`IdleState.READER_IDLE` will close a connection if logstash takes a long time to process the current batch.


ph:
If Logstash takes a long time to process a batch, the **IdleStateHandler** will trigger a `IdleState.READER_IDLE` event and logstash will close the current connection.

We need to check if Logstash is currently processing a batch and implement the following logic.

- If we aren't processing => **close**
- If are => **ignore** that event.

PR #142, opened 2016-09-22
Dont close the connection when the `IdleState.READER_IDLE` occur when processing a batch.

ph:
When logstash takes more time to process a batch than the `client_inactivity_timeout`,
the server was closing the connection in the middle of working on a batch. This commit changes the behavior,
if logstash is processing a batch and the event occur we wont drop the connection otherwise we will close it.

Fixes: #141, #143


ph:
@talevy would you mind taking a look at this, since you already know this part of the code?

ph:
The failling travis seems to be a timing issue, with how the integration test make the assertions with the external process.

talevy:
@ph I have a question about the keepAlive. what happens if there are no messages to be processed and we close the connection? will filebeat properly handle the networking in this scenario?

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -84,8 +84,10 @@ private void writeAck(ChannelHandlerContext ctx, byte protocol, int sequence) {
     }
 
     private void clientTimeout() {
-        logger.debug("Client Timeout");
-        this.context.close();
+        if(!processing.get()) {
```
talevy: do we have tests for this timeout?

ph:
> I have a question about the keepAlive. what happens if there are no messages to be processed and we close the connection? will filebeat properly handle the networking in this scenario?

They will get a EOF, and it will reconnect maybe we should change the default values for 15 minutes instead of seconds? what do you think?.

src/main/java/org/logstash/beats/BeatsHandler.java
ph: We have test to cover that will check that a connection is close when the timeout is reached its in the `ServerTest`

talevy:
> maybe we should change the default values for 15 minutes instead of seconds? what do you think?.

maybe default to 1 minute?

ph:
Will make that change to 1 minute.

talevy:
LGTM

PR #145, opened 2016-09-23
rFix an issue when a rejectedError was raise by the executor

ph:
The problem was the order of the shutdown was not correctly defined and
was async, this PR change the flow to wait for the workgroup and
bossgroup executor to be closed before terminating the idlestate
executor.

Also, I've changed the parameters of shutdownGracefully to use the
default method of netty which users saner default values for the quiet
time and make the code close more reliably when shutting down
Logstash.


ph:
the flaky test are fixed in #142

talevy:
LGTM

Issue #146, opened 2016-09-26
The host setting seems to not have any effect


neuroid:
- Version: 3.1.3 (bundled with logstash 2.4.0)
- Operating System: Debian Jessie

My config looks like this:

```
input {
  beats {
    host => "127.0.0.1"
    port => 5044
  }
}
```

Running with `--debug` shows:

```
[snip]
config LogStash::Inputs::Beats/@host = "127.0.0.1" {:level=>:debug, :file=>"logstash/config/mixin.rb", :line=>"154", :method=>"config_init"}
config LogStash::Inputs::Beats/@port = 5044 {:level=>:debug, :file=>"logstash/config/mixin.rb", :line=>"154", :method=>"config_init"}
[snip]
Beats inputs: Starting input listener {:address=>"127.0.0.1:5044", :level=>:info, :file=>"logstash/inputs/beats.rb", :line=>"153", :method=>"register"}
[snip]
```

However, when I check the listening sockets I get:

```
# netstat -pan | grep java | grep LISTEN
tcp        0      0 0.0.0.0:5044            0.0.0.0:*               LISTEN      18247/java
```

ph:
I Will take a look

ph:
I can confirm its a bug, 

instead of `bind(port)` it should use `bind(String inetHost, int inetPort)`

bkuberek:
Sorry to poke this thread but I am having an issue with the beats plugin not listening on port 5044 as well. Is this bug affecting version 2.3.0 (Mar 29, 2016) ?

Thanks

ph:
This issue was about 2.4.0 which included the java rewrite of the beat input.

@bkuberek 2.3.0 was shipped with an older version of beats which was written in ruby at that time, 3.0.4 ? or something close.

by looking at the code the server accepts an address: https://github.com/logstash-plugins/logstash-input-beats/blob/v3.0.4/lib/lumberjack/beats/server.rb#L51  and the input pass that value to the library.

So I think it should be fine.

You can connect that port? Anything in the log?

bkuberek:
@ph Turns out we had a syntax error in the logstash configuration and the service kept restarting. After we fixed that it's all good. Thank you for taking the time to answer my question. I really appreciate it.

Best,

Bastian

PR #148, opened 2016-09-28
Allow the beats inputs to bind to a specific host

ph:
The host options was not correctly send to the server code.

Fixes: #146

### DIFF
--- a/spec/support/shared_examples.rb
+++ b/spec/support/shared_examples.rb
@@ -25,7 +25,7 @@ shared_examples "Common Event Transformation" do
     expect(subject.get("tags")).to include(tag)
   end
 
-  context "when the `beast.hotname` doesnt exist on the event" do
+  context "when the `beats.hostname` doesnt exist on the event" do
     let(:already_exist) { "already_exist" }
     let(:event_map) { super.merge({ "host" => already_exist }) }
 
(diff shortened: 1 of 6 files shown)


spec/support/shared_examples.rb
```diff
@@ -25,7 +25,7 @@
     expect(subject.get("tags")).to include(tag)
   end
 
-  context "when the `beast.hotname` doesnt exist on the event" do
+  context "when the `beast.hostname` doesnt exist on the event" do
```
talevy: s/beast/beats

src/main/java/org/logstash/beats/Server.java
```diff
@@ -29,21 +29,27 @@
...
     private final int port;
     private final NioEventLoopGroup bossGroup;
     private final NioEventLoopGroup workGroup;
+    private final String host;
     private IMessageListener messageListener = new MessageListener();
     private SslSimpleBuilder sslBuilder;
 
     private final int clientInactivityTimeoutSeconds;
 
     public Server(int p) {
```
talevy: is this constructor ever used? maybe we can go so far as to force an explicit host at all times and fall back to the `@host` used in the logstash plugin config

ph: true, I will remove it for simplicity sake.

ph:
I've made the requested changes.

src/main/java/org/logstash/beats/Server.java
```diff
@@ -29,21 +29,23 @@
 
     static final long SHUTDOWN_TIMEOUT_SECONDS = 10;
     private static final int DEFAULT_CLIENT_TIMEOUT_SECONDS = 15;
-
+    private static final String DEFAULT_HOST = "0.0.0.0";
```
talevy: not used

ph: done

talevy:
LGTM

Issue #153, opened 2016-10-28
bug: ssl_certificate_authorities setting only loads the first certificate in each file.


jordansissel:
Buggy code linked here: https://github.com/logstash-plugins/logstash-input-beats/blob/776f05ce52111d39e394798428598abfd4dbbf68/src/main/java/org/logstash/netty/SslSimpleBuilder.java#L152

```
in = new FileInputStream(certificate);
collections[i] = (X509Certificate) certificateFactory.generateCertificate(in);
```

CertificateFactory.generateCertificate(...) only returns a single X509Certificate. However, a single file can contain multiple certificates.

Fix: Use CertificateFactory.generateCertificates(...) (the plural method name) which returns an array of X509Certificate.

ph:
lets make sure we also update the doc to reflect that changes.

This sentence under the **certificates_authorities** need to be removed.

> This feature only supports certificates that are directly signed by your root CA. Intermediate CA are currently not supported.

PR #154, opened 2016-10-28
Fix an issue with when client authentification is on and CA contains multiples CA

ph:
Fix an issue when using client authentification and multiple CA defined
in the same **certificate_authorities** files. When the server read the
file it would only pick the first certificate in the file and ignore the
rest.

This problem could result in `unable to find valid certification path to
requested target` in Logstash and would refuse to let bet connect to it.

Fixes: #153


src/main/java/org/logstash/netty/SslSimpleBuilder.java
```diff
@@ -138,26 +141,20 @@ public SslHandler build(ByteBufAllocator bufferAllocator) throws IOException, No
...
-            InputStream in = null;
-
-            try {
-                in = new FileInputStream(certificate);
-                collections[i] = (X509Certificate) certificateFactory.generateCertificate(in);
-            } finally {
-                if(in != null) {
-                    in.close();
-                }
+            try(InputStream in = new FileInputStream(certificate)) {
```
jordansissel: Can you add `logger.debug("Loading certificates from file {}", certificate)` ?

jordansissel:
LGTM.

I tested manually also to verify that the new spec failed without this PR. :white_check_mark:

src/main/java/org/logstash/netty/SslSimpleBuilder.java
ph: certainly!

Issue #151, opened 2016-10-26
Validate the presence of `ssl_certificates_authorities` when verify mode is set to FORCE_PEER.


ph:
Validate the presence of `ssl_certificates_authorities` when verify mode is set to FORCE_PEER and throw an error if not present.

PR #155, opened 2016-10-28
Validate the presence of certificate authorities

ph:
When we set the `ssl_verify_mode` to `peer` or `force_peer` we should
make the `certificate_authorities` mandatory, when the option is not set
we will now throw a configuration error.

Fixes: #151


talevy:
LGTM

PR #162, opened 2016-11-23
Remove restriction about intermediate CAs not being supported

dedemorton:
I believe that this restriction should have been removed in PR https://github.com/logstash-plugins/logstash-input-beats/pull/154 but wasn't. 

@ph Please review.


ph:
LGTM

Issue #163, opened 2016-11-23
Connection closed / logstash: Client Timeout / filebeat: EOF


jautz:
## Summary

In my experiments with the very promising filebeat/logstash setup for remote logging I ran into an issue with connections being closed prematurely. That is, when producing log data in intervals greater than five seconds the connection between filebeat and logstash would be closed. I did not witness any data loss but at least data duplication (only once so far).

Still I wanted to increase the timeout to something like a minute. I was able to locate logstash's "Client Timeout" debug message in source file `src/main/java/org/logstash/beats/BeatsHandler.java`.

I found the [FileBeat EOF Error discussion](https://discuss.elastic.co/t/filebeat-eof-error/60882) and the (slightly related) [logstash-plugins/logstash-input-beats@02bdcca commit](https://github.com/logstash-plugins/logstash-input-beats/commit/02bdccaf9bc6e86a2ba1517b5aaf6ab524c12d07) that mentions the `client_inactivity_timeout` setting that has been introduced with [Pull Request #108](https://github.com/logstash-plugins/logstash-input-beats/pull/108).

But the bottom line is: I failed to make it work. Either my increased timeout is simply ignored or I am looking in the wrong places. Any hints from your side will be highly appreciated.

So here is my setup and the reproducible steps that provoke error messages. Both machines (filebeat client and logstash server) run Debian Jessie.

## Server: logstash process accepting beats

```
$ apt-cache policy logstash
logstash:
  Installed: 1:5.0.1-1
  Candidate: 1:5.0.1-1
  Version table:
 *** 1:5.0.1-1 0
        500 https://artifacts.elastic.co/packages/5.x/apt/ stable/main amd64 Packages
        100 /var/lib/dpkg/status


$ cat /etc/logstash/pipeline.conf
input {
  beats {
    port => 5044
    # close idle connection after 1m (default: 15s)
    client_inactivity_timeout => 60
  }
}

output {
  file {
    path => "/var/log/logstash/timeout_demo.log"
  }
}


$ sudo /usr/share/logstash/bin/logstash --path.settings /etc/logstash -f /etc/logstash/pipeline.conf --log.level=debug
Sending Logstash's logs to /var/log/logstash which is now configured via log4j2.properties
```

## Client: filebeat process producing beats

```
$ apt-cache policy filebeat 
filebeat:
  Installed: 5.0.1
  Candidate: 5.0.1
  Version table:
 *** 5.0.1 0
[...]

PR #164, opened 2016-11-23
Correctly send the `client_inactivity_timeout` value to the beats server

ph:
The `client_inactivity_timeout` wasn't not correctly send to the Server
class so the server was using a default value instead. This PR add an
assertions on the value that the server class should receive, we have a
unit test in the java code that test the behavior of the timeout.

Fixes: #163


colinsurprenant:
LGTM

PR #165, opened 2016-11-23
Mark the congestion threshold as deprecated

ph:
This option is not used anymore and should have been marked as
deprecated, the input now uses at keep alive to FB (sending ack 0) to
inform the client that some back pressure is happening.

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -118,7 +118,7 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   # The number of seconds before we raise a timeout. 
   # This option is useful to control how much time to wait if something is blocking the pipeline.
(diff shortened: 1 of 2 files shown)


lib/logstash/inputs/beats.rb
```diff
@@ -118,7 +118,7 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   # The number of seconds before we raise a timeout. 
   # This option is useful to control how much time to wait if something is blocking the pipeline.
-  config :congestion_threshold, :validate => :number, :default => 5
+  config :congestion_threshold, :validate => :number, :default => 5, :deprecated => "This option is now deprecated, the plugin now use a keep alive mechanism to inform Filebeat"
```
jsvd: I think we should maybe inform that this is now done automatically, like:

`This option is now obsolete since congestion control is done automatically`?

ph: make sense will change it.

ph:
updated with your suggestion.

Issue #132, opened 2016-09-16
log4j warning when run from 2.4


ph:
When you start the plugin in 2.4 we get a warning.

```
rclarke@es-rclarke:100456$ ~/elastic/stack/logstash-2.4.0/bin/logstash agent --config ./logstash.conf --log ./logstash.log --quiet
Sending logstash logs to ./logstash.log.
Settings: Default pipeline workers: 4
log4j:WARN No appenders could be found for logger (io.netty.util.internal.logging.InternalLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

ph:
I believe #130 fixed the issue

ph:
confirmed that the #130 PR fixed that issue

robin13:
It seems that this has come back in version 3.1.8 of the plugin (deployed with Logstash 2.4.1)

```
input {
    beats {
        port => 12345
    }
}

output {
    stdout {
    }
}
```

results in

```
log4j:WARN No appenders could be found for logger (io.netty.util.internal.logging.InternalLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Settings: Default pipeline workers: 4
```

ph:
confirmed its back :(

```
ph@sashimi~/Downloads/logstash-2.4.1 bin/logstash -e 'input { beats { port => 6000 }}'
	log4j:WARN No appenders could be found for logger (io.netty.util.internal.logging.InternalLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Settings: Default pipeline workers: 4
Pipeline main started
```

renevdm:
It seems like the error is back again:

`# service logstash configtest
log4j:WARN No appenders could be found for logger (io.netty.util.internal.logging.InternalLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Configuration OK
`
No events flow into elasticsearch via logstash. But when I disable all beats inputs, the error messages disappear and events do flow into elasticsearch.

ph:
[...]

PR #166, opened 2016-11-24
Correctly initialize the logger for the netty code

ph:
Since the logger is configured as a class variable in Netty, the error
was produced when the classes were required and before the logger was
correctly setup for LS 2.4. This PR change when the java_import is done
to make sure the log4j's appender are correctly setup.

Fixes: #132

It appear that #130 didn't completely fixes that problem or something changed between Logstash versions, I haven't checked that.

Also I don't think there is an easy way to assert for this in a test? I did manual testing on both 2.4.x and 5.0 and the logger seems to works as expected.


ph:
@robin13 Can you test this fix?

talevy:
after rebase, LGTM

PR #169, opened 2016-11-29
Uses SO_LINGER on the socket option

ph:
When a idle client is detected the server will close the connection, if
you look at the TCP State the following scenario will happen.

- Server is set to FIN_WAIT_2
- Client is set to CLOSE_WAIT

Since FB never read from the socket it wont complete the handshake to
close the connection, change the socket option for `SO_LINGER` will
force the connection to be terminated from the server side.

On the FB side, the client will still think its connected until he need
to write something to the wire, when it happen it will just reconnect
to the server.

Using SO_LINGER is necessary in that case, because we have no way to
communicate to FB client that his connection was idle for too long, this
kind of two way messaging is not implemented yet in the protocol level.

Related to https://github.com/elastic/logstash/issues/6300


jordansissel:
> Since FB never read from the socket

Is this true? Doesn't libbeat use null acks (for example) as a heartbeat?

jordansissel:
Additionally, isn't SO_LINGER off by default on linux?

jordansissel:
Oh neat. SO_LINGER is enabled on JRuby (Java?) by default it seems:

```
% rvm 1.7.26,default do ruby -rsocket -e 'p TCPSocket.new("localhost", 22).getsockopt(:SOCKET, :LINGER)'
#<Socket::Option INET SOCKET LINGER on>
#<Socket::Option: INET6 SOCKET LINGER off 0sec>
```

src/main/java/org/logstash/beats/Server.java
```diff
@@ -1,10 +1,7 @@
 package org.logstash.beats;
 
 import io.netty.bootstrap.ServerBootstrap;
-import io.netty.channel.Channel;
-import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInitializer;
-import io.netty.channel.ChannelPipeline;
+import io.netty.channel.*;
```
jordansissel: I prefer not using wildcard imports, but wont' block the PR on this.

jordansissel:
Code LGTM, but I haven't tested.

ph:
@jordansissel I was as surprised as you that SO_LINGER is not by default on jruby, I would have think to be off :)

ph:
@jordansissel IIRC it only do health check with the acks when the library need to send something, I don't think it does when there no event to push.

Issue #174, opened 2016-12-22
LS 5.1.1 beats input plugin unsupported cipher suite TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384


buzzdeee:
I'm testing LS 5.1.1 (ES/Kibana are already at that version)

Because of this issue 
https://github.com/elastic/logstash/issues/6121
I had this in my beats input definition:

```
  beats {
    port => 12344
    ssl => true
    ssl_certificate => "/etc/logstash/logstash.crt"
    ssl_key => "/etc/logstash/logstash.key"
    ssl_certificate_authorities => [ "/etc/logstash/logstash-ca.crt" ]
    ssl_verify_mode => "force_peer"
    cipher_suites => [ 'TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256' ]
    type => "logs"
  }
```

that was working properly with LS 2.4.1.
Then on startup of LS 5.1.1 I got:

```
[2016-12-21T23:36:25,848][INFO ][logstash.inputs.beats    ] Beats inputs: Starting input listener {:address=>"0.0.0.0:12344"}
[2016-12-21T23:36:26,400][ERROR][logstash.agent           ] Pipeline aborted due to error {:exception=>#<LogStash::ConfigurationError: Cipher `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` is not available>, :backtrace=>["/usr/local/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-3.1.12-java/lib/logstash/inputs/beats.rb:184:in `create_server'", "/usr/local/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-beats-3.1.12-java/lib/logstash/inputs/beats.rb:172:in `register'", "/usr/local/logstash/logstash-core/lib/logstash/pipeline.rb:353:in `start_inputs'", "org/jruby/RubyArray.java:1613:in `each'", "/usr/local/logstash/logstash-core/lib/logstash/pipeline.rb:352:in `start_inputs'", "/usr/local/logstash/logstash-core/lib/logstash/pipeline.rb:228:in `start_workers'", "/usr/local/logstash/logstash-core/lib/logstash/pipeline.rb:183:in `run'", "/usr/local/logstash/logstash-core/lib/logstash/agent.rb:292:in `start_pipeline'"]}
[...]

PR #181, opened 2017-01-31
Bootstrap check to see if the configured temporary directory is mounted with noexec

ph:
When you start logstash and the temporary directory is mounted with
`noexec` and you are using SSL this will make the netty library fails to
load the native library. This PR add a bootstrap check and a betters
message than missing a ciphers.

Fixes: #174

Notes for reviewer: Writing test for this isn't an easy task, I've tested it manually by creating a ramdisk with the `noexec` option and changing the config/logstash.yml directory to point to that.

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -5,6 +5,7 @@ require "logstash/timestamp"
 require "logstash/codecs/identity_map_codec"
 require "logstash/codecs/multiline"
 require "logstash/util"
+require "logstash/errors"
 require "logstash-input-beats_jars"
 
 # This input plugin enables Logstash to receive events from the
@@ -60,6 +61,7 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   require "logstash/inputs/beats/decoded_event_transform"
   require "logstash/inputs/beats/raw_event_transform"
   require "logstash/inputs/beats/message_listener"
+  require "logstash/inputs/beats/stats_helper"
   require "logstash/inputs/beats/tls"
 
   config_name "beats"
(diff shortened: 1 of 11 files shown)


lib/logstash/inputs/beats/stats_helper.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      parts = path.split(::File::SEPARATOR)
+      number_of_iteration = parts.size
+
+      results = {}
+
+      number_of_iteration.times.collect do
+        new_path = ::File.join(*parts)
+
+        # we are interested all all the hierachy
```
ph: double all.. :bummer.

lib/logstash/inputs/beats.rb
```diff
@@ -175,6 +177,7 @@ def register
   def create_server
     server = org.logstash.beats.Server.new(@host, @port, @client_inactivity_timeout)
     if @ssl
+      check_no_exec_mount!
```
jordansissel: Also, Netty has this https://netty.io/4.0/api/io/netty/handler/ssl/OpenSsl.html#isAvailable()

We can check OpenSsl.isAvailable(), or if we want the exception, call `OpenSsl.ensureAvailability()`. I have tried this in `tealess` and it fails when /tmp is noexec (or if the .so library fails to load in any way)

ph: woa,  that make the code a lot simplier I wasn't aware of that method, I guess I've looked at the wrong place.

jordansissel:
@ph did you still want to move this to use OpenSsl.isAvailable() ?

ph:
Yes, it will be much better and simpler

jordansissel:
I think we are agreed to move away from tcnative/ssl and toward Java ssl -- https://github.com/logstash-plugins/logstash-input-beats/issues/188

Closing this in favor of #188

PR #182, opened 2017-01-31
Simple fix to make the input not set the CAs when the verifymode is s…

nick-george:
For a bit of context, please see the discussion (or lack thereof) at https://discuss.elastic.co/t/beats-input-not-behaving-as-documented/73360.

I attempted to run (and possibly write new) tests for this, but got stuck with some errors running 'bundle exec rspec'. Any help in this area would be much appreciated. 

Cheers, 
Nick


breml:
@nick-george I stumbled over this issue (and the missleading documentation already mentioned in #196) today. Thanks for your PR, I like it.

PR #189, opened 2017-02-13
Properly close multiple idle connections

masaruh:
Fixes an issue where it doesn't close connections even after client_inactivity_timeout
has passed when there are multiple connections.

### DIFF
--- a/src/test/java/org/logstash/beats/ServerTest.java
+++ b/src/test/java/org/logstash/beats/ServerTest.java
@@ -1,23 +1,31 @@
 package org.logstash.beats;
 
 import io.netty.bootstrap.Bootstrap;
-import io.netty.channel.*;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.net.ConnectException;
 import java.util.Collections;
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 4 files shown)


ph:
Fix the test run I've applied this patch and this should fix any travis errors, we did some changes upstream that make the load order more important.

```diff
diff --git a/spec/inputs/beats/codec_callback_listener_spec.rb b/spec/inputs/beats/codec_callback_listener_spec.rb
index 5c08b57..9329f04 100644
--- a/spec/inputs/beats/codec_callback_listener_spec.rb
+++ b/spec/inputs/beats/codec_callback_listener_spec.rb
@@ -1,7 +1,7 @@
 # encoding: utf-8
+require_relative "../../spec_helper"
 require "logstash/inputs/beats"
 require "logstash/event"
-require "spec_helper"
 require "thread"
 
 describe LogStash::Inputs::Beats::CodecCallbackListener do
diff --git a/spec/spec_helper.rb b/spec/spec_helper.rb
index 67e3a42..7ec54aa 100644
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,8 +1,8 @@
 # encoding: utf-8
+require "logstash/devutils/rspec/spec_helper"
 require "rspec"
 require "rspec/mocks"
 require "rspec/wait"
-require "logstash/devutils/rspec/spec_helper"
 require "logstash/codecs/plain"
 require_relative "support/logstash_test"
 require_relative "support/helpers"

```

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -52,18 +94,20 @@ public void run() {
...
-
         try {
             Long started = System.currentTimeMillis() / 1000L;
 
-            connectClient().sync().channel().closeFuture().sync();
+
+            for (int i = 0; i < concurrentConnections; i++) {
+                connectClient();
+            }
+            latch.await(10, TimeUnit.SECONDS);
```
ph: TIL

ph:
@masaruh I've tested this PR and I still see connection in the netstat, long after the timeout period.
I wonder if something else prevent the timeout event to be executed.

masaruh:
Hmm, it works for me... Making it not sharable should work too.

Since #191 supersedes this issue, I'm going to close this issue. But pushing new change which makes it not have state, just in case...

ph:
@masaruh I agree I would prefer not having to keep a state, I have no other indication that the queue is blocked at the moment and this plugin need to work in both 2.x 5.x :(

Issue #185, opened 2017-02-04
Investigate LS->beats read timeout


urso:
For all general issues, please provide the following details for fast resolution:

- Version: beats and logstash 5.2

We've got a number of reports about beats->Logstash connection loss due to timeout:

```
ERR Failed to publish events caused by: read tcp : i/o timeout
```

This timeout occurs when beats is waiting for ACK from the input plugin. In case logstash is busy, a heartbeat signal `ACK 0` is supposed to be send every 5 seconds, to signal beats the batch of events being send are still in progress. Plus beats employ a read timeout of 30 seconds. That is, we missed about 6 heartbeat signals (the timeout in beats is configurable).

We should make the `heartbeat` interval configurable and have some integration tests with beats or [go-lumber](https://github.com/elastic/go-lumber) (go-lumber allows to wait for partial ACKs, so we can verify heartbeat ACKs being received).

Interestingly, in all cases users did report LS being busy. e.g. in one incident user did report logstash having very high resource using trying to send events to failing ES cluster over some long period of time. [This one](https://discuss.elastic.co/t/filebeat-duplicate-events/73693/3) has some 'grok timeout' error messages.

Will add more reports coming in (or old ones if I find them) to this ticket.

urso:
Another one: https://discuss.elastic.co/t/metricbeats-timeout/73890

ph:
This is revealling to be an easy fix, the `BeatsHandler` is marked as shareable but has states in it.

ph:
I think #191 also address this problem, will do a bit more testing to make sure it does.

ph:
fixed in #191

Issue #184, opened 2017-02-04
Issue with BeatsHandler.java


fbacchella:
Reading the code of the BeatsHandler, I see many things that I don't understand.

It's marked Sharable, but there is state in it, with the variables `processing` and `context`.

With:

    logger.debug("Sending a new message for the listener, sequence: " + message.getSequence());

A debug logger should not do operation without first checking if it's needed, especially when it's call frequently, like this one which is called for every message.

ph:
Agree, I will change that.

Issue #178, opened 2017-01-19
Logstash not closing close_wait connections.


sc-chad:
- Version: 
logstash 5.1.2 
logstash-input-filebeat: 3.1.12
- Operating System: Amazon Linux
- Config File (if you have sensitive info, please remove it): 
- Sample Data:
- Steps to Reproduce:

Similar issue to: https://github.com/elastic/logstash/issues/6300
Previous discussed issue: https://github.com/logstash-plugins/logstash-output-s3/issues/120

Issue: I believe we have narrowed it down to the filebeat plugin not closing all the connections when using an Amazon ELB.  I have tried lowering the default timeout to 60 seconds which didn't seem to work.

Example:
```
java      15038      logstash 2152u     IPv4         1449165042         0t0        TCP prod-graylog-processor-775e2946:lxi-evntsvc->ip-10-0-5-216.ec2.internal:16700 (CLOSE_WAIT)
java      15038      logstash 2153u     IPv4         1449165044         0t0        TCP prod-graylog-processor-775e2946:lxi-evntsvc->ip-10-0-5-216.ec2.internal:16702 (CLOSE_WAIT)
java      15038      logstash 2154u     IPv4         1449165046         0t0        TCP prod-graylog-processor-775e2946:lxi-evntsvc->ip-10-0-1-185.ec2.internal:14513 (CLOSE_WAIT)
```

Showing command and count: 
```
/usr/sbin/lsof | grep "logstash" | grep "CLOSE_WAIT" | wc -l
2173
```

feichaohao:
How to solve the problem?

feichaohao:
filebeat  set : 
spool_size: 1024
 idle_timeout: 5s

logstash set : 
beat{         port => 5044
        type => "log"
        add_field=>{"short_message"=>"product"}
        add_field=>{"host"=>"host"}
    }

ph:
fixed in #191

patsevanton:
Hello!
in which version of Logstash is this bug fixed?

PR #191, opened 2017-02-13
Client timeout issue, and debug/trace statements refactor

ph:
This PR changes the following:

- It make the beats handler not `shareable`, to make sure the processing context is only related to the current channel handler, this will make the  **client_timeout** and **keep alive** check  done with the right context. Use @masaruh unit test from #189

- Only record debug statement when we are actually in debug mode

- Use trace instead of debug in the beats parser to make the logging less noisy.

- Make sure the test suite run on travis by changing the requires ordering.

Fixes: #185, #184, #178,


ph:
@masaruh Can you test this PR? I've used your unit test from #189 but I've made the handler not shareable.

masaruh:
Yes, it works as expected.

ph:
@talevy Can you take a look at the code?

spec/inputs/beats/codec_callback_listener_spec.rb
```diff
@@ -1,4 +1,5 @@
 # encoding: utf-8
+require_relative "../../spec_helper"
```
talevy: what is the other `spec_helper` on line 5?

spec/spec_helper.rb
```diff
@@ -1,8 +1,8 @@
 # encoding: utf-8
+require "logstash/devutils/rspec/spec_helper"
```
talevy: why the re-org here?

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -39,7 +38,8 @@ public void channelRead0(ChannelHandlerContext ctx, Batch batch) throws Exceptio
         processing.compareAndSet(false, true);
 
         for(Message message : batch.getMessages()) {
-            logger.debug("Sending a new message for the listener, sequence: " + message.getSequence());
+            if(logger.isDebugEnabled())
```
talevy: can we add `{` `}` around this?

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -228,7 +229,8 @@ private void transition(States next) {
     }
 
     private void transition(States nextState, int requiredBytes) {
-        logger.debug("Transition, from: " + currentState + ", to: " +  nextState + ", requiring " + requiredBytes + " bytes");
+        if(logger.isTraceEnabled())
```
talevy: nitpick, same as above

spec/spec_helper.rb
ph: If we dont do this, the test will fail complaining it cannot find log4j.

spec/inputs/beats/codec_callback_listener_spec.rb
ph: we should not have that, I will fix it.

spec/spec_helper.rb
talevy: ah

src/main/java/org/logstash/beats/BeatsParser.java
```diff
@@ -207,7 +207,8 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
                 batch.addMessage(message);
 
                 if(batch.size() == batch.getBatchSize()) {
-                    logger.debug("Sending batch size: " + this.batch.size() + ", windowSize: " + batch.getBatchSize() +  " , seq: " + sequence);
+                    if(logger.isTraceEnabled())
```
talevy: nitpick, same as above

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -34,9 +42,43 @@ public void setUp() {
...
+        final Random random = new Random();
+
+        final CountDownLatch latch = new CountDownLatch(concurrentConnections);
 
         final Server server = new Server(host, randomPort, inactivityTime);
+        final AtomicBoolean otherCause = new AtomicBoolean(false);
+        server.setMessageListener(new MessageListener() {
+            public void onNewConnection(ChannelHandlerContext ctx) {
+                // Make sure connection is closed on exception too.
+                if (random.nextInt(10) < 1) {
```
talevy: I'm a little confused, we are connecting 10 clients, with a 10% chance of throwing an exception (which we are expecting), isn't this pretty likely not to fire?

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -34,9 +42,43 @@ public void setUp() {
...
+            }
+
+            @Override
+            public void onConnectionClose(ChannelHandlerContext ctx) {
+                latch.countDown();
+            }
+
+            @Override
+            public void onNewMessage(ChannelHandlerContext ctx, Message message) {
+                // Make sure connection is closed on exception too.
```
talevy: what side effect exists if a connection is not closed? I don't see anything explicit in the assertions in this test?

talevy: oh, nevermind. that is what the `latch` is for

src/test/java/org/logstash/beats/ServerTest.java
ph: I think the case here is to make sure that the connection is closed indepent of a timeout or an exception, I guess they could be 2 different tests WDYT?

talevy: hmm, I think it is fine how it is, since it is testing cases for closing. I just think maybe this probability should be increased?

masaruh: If the probability is high enough, all connection can get exception. In that case, no connection waits for defined timeout and assertion that checks it would fail.
We might want to create another test for error cases.

ph: I will create a new test.

ph:
@talevy @masaruh  I've updated this PR with your comments

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -52,18 +92,74 @@ public void run() {
...
 
         try {
             Long started = System.currentTimeMillis() / 1000L;
 
-            connectClient().sync().channel().closeFuture().sync();
+
+            for (int i = 0; i < concurrentConnections; i++) {
+                connectClient();
+            }
+            assertThat(latch.await(10, TimeUnit.SECONDS), is(true));
```
talevy: you can just use `assertTrue` here?

src/test/java/org/logstash/beats/ServerTest.java
```diff
@@ -52,18 +92,74 @@ public void run() {
...
 
             Long ended = System.currentTimeMillis() / 1000L;
 
             double diff = ended - started;
             assertThat(diff, is(closeTo(inactivityTime, 0.1)));
-
+            assertThat(exceptionClose.get(), is(false));
```
talevy: use assertTrue?

talevy:
changes LGTM, just small nitpick about the `is(true)` and `is(false)`. maybe `assertFalse` and `assertTrue` would be clearer? 

I know, sometimes the old pattern is better  :)

Issue #192, opened 2017-02-13
Include the Beats source host/ip in Exception message


PhaedrusTheGreek:
When receiving a message like the following:

```
[2017-02-09T12:41:33,571][ERROR][org.logstash.beats.BeatsHandler] Exception: not an SSL/TLS record: < blah >
```

It would be helpful if we could log the source IP address / hostname of the connecting Beats agent.

PR #198, opened 2017-04-10
Add information concerning the remote host

ph:
This PR add remote host information if available when an exception occur

Fixes : #192


src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -54,8 +56,14 @@ public void channelRead0(ChannelHandlerContext ctx, Batch batch) throws Exceptio
 
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        InetSocketAddress remoteAddress = (InetSocketAddress) ctx.channel().remoteAddress();
+
+        if (remoteAddress != null) {
+            logger.error("Exception: " + cause.getMessage() + ", from: " + remoteAddress.getHostString() + ":" + remoteAddress.getPort());
```
jordansissel: InetSocketAddress has a reasonable toString you can use for the 'from' part

```
>> java.net.InetSocketAddress.new("foo", 2345).toString
=> "foo:2345"
```

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -54,8 +56,14 @@ public void channelRead0(ChannelHandlerContext ctx, Batch batch) throws Exceptio
 
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        InetSocketAddress remoteAddress = (InetSocketAddress) ctx.channel().remoteAddress();
```
jordansissel: Does this work in all exceptions? Linux's getpeername will fail on sockets that have been closed,.

jordansissel: Ahh, nevermind. Java's SOcketChannel.getRemoteAddress can throw exceptions, but Netty's Channel.remoteAddress() will not (it returns null instead)

src/main/java/org/logstash/beats/BeatsHandler.java
ph: Do you know if it will do a dns lookup? `getHostString()` wont.

ph:
@jordansissel updated with your changes.

jordansissel:
I'm checking on the dns thing. Waiting on mercurial to checkout jdk again ☹️

jordansissel:
For posterity, InetSocketAddress.toString() is this:

```
        public String toString() {
            if (isUnresolved()) {
                return hostname + ":" + port;
            } else {
                return addr.toString() + ":" + port;
            }
        }
```

Looks ok :)

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -54,8 +56,14 @@ public void channelRead0(ChannelHandlerContext ctx, Batch batch) throws Exceptio
...
+
+        if (remoteAddress != null) {
+            logger.error("Exception: " + cause.getMessage() + ", from: " + remoteAddress.toString());
```
jordansissel: Are all exceptions requiring operator action? If so, ERROR log level makes sense. However, I have a suspicion that most network problems will not require operator action to resolve, so maybe this should be `logger.warn` or `logger.info` ?

My thought is that the network is assumed to be unstable, so temporary problems will result in exceptions but otherwise resolve without human intervention, and these temporary problems should not be logged as error or warn.

ph: I think you are right here, It would be OK to use `.info`, error level might just create more unnecessary issues than we actually require.

ph:
updated with the appropriate log level.

jordansissel:
LGTM

PR #200, opened 2017-04-13
implement current and peak connection count metrics

jsvd:
add metrics in the context of https://github.com/elastic/logstash/issues/6831

### DIFF
--- a/lib/logstash/inputs/beats/message_listener.rb
+++ b/lib/logstash/inputs/beats/message_listener.rb
@@ -19,6 +19,8 @@ module LogStash module Inputs class Beats
       @queue = queue
       @logger = input.logger
       @input = input
+      @metric = @input.metric
+      @peak_connection_count = Concurrent::AtomicFixnum.new(0)
 
       @nocodec_transformer = RawEventTransform.new(@input)
       @codec_transformer = DecodedEventTransform.new(@input)
@@ -44,10 +46,12 @@ module LogStash module Inputs class Beats
 
     def onNewConnection(ctx)
       register_connection(ctx)
+      increment_connection_count()
     end
 
     def onConnectionClose(ctx)
       unregister_connection(ctx)
+      decrement_connection_count()
     end
 
     def onChannelInitializeException(ctx, cause)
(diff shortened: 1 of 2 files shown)


jsvd:
@ph can I get your input in the strategy used here? since we don't have a metric that keeps the highest value I had to implement it here with two atomic fixnums. wdyt?

Also, I just noticed we don't add the type of the plugin to input plugin metrics:

```
% curl -s localhost:9600/_node/stats | jq ".pipeline.plugins.inputs"
[
  {
    "id": "88fd96b845d7b241ee7308afcf9999a32e1ddf0f-1",
    "current_connections": 0,
    "peak_connections": 3
  }
]
```

I'll submit a PR in core for that

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -19,6 +19,9 @@ def initialize(queue, input)
       @queue = queue
       @logger = input.logger
       @input = input
+      @metric = @input.metric
+      @peak_connection_count = Concurrent::AtomicFixnum.new(0)
+      @current_connection_count = Concurrent::AtomicFixnum.new(0)
```
ph: I think we can derive the value of this by doing `@connections_list.size`, since we are using a *ThreadSafe::Hash*, the `#size` should also be synchronized.

ph: I am +1 for the strategy of the peak connection.

jsvd:
@ph remark addressed 👍

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -107,5 +111,18 @@ def from_filebeat?(hash)
...
     end
+
+    def increment_connection_count
+      current_connection_count = @connections_list.size
+      @metric.gauge(:current_connections, current_connection_count)
+      if current_connection_count > @peak_connection_count.value
+        @peak_connection_count.value = current_connection_count
+        @metric.gauge(:peak_connections, @peak_connection_count.value)
+      end
+    end
```
ph: We can add a simple unit test for this? Since the `MessageListener` is self contained?

jsvd:
@ph added metrics to metric calculation

Issue #199, opened 2017-04-12
Add warning when using beats input with multiline when multiple hosts are configured


ppf2:
This is a high priority doc issue.

Here is the scenario:

* Beats is configured to have output.logstash output with multiple hosts in the hosts list.
* Logstash pipeline has a beats input with multiline codec in place to perform multiline processing.

The problem with the above setup is that if filebeat does not know about multiline (i.e. multiline is not configured via filebeat but as part of a LS input - this does not only apply to LS beats input, but any LS input that supports multiple hosts), it does not know which events belong together so it will happen that parts are sent to one host and other parts to an other host, causing a mixing of streams and corrupted event data generated.

Per @ruflin This is a known limitation and there isn't really a solution other than to configure multiline in beats instead of in Logstash.  

It is best if we can document this on both sides, not just in Beats documentation.  So this is the doc ticket for LS.  I will suggest calling this out in the managing multiline section (https://www.elastic.co/guide/en/logstash/current/multiline.html) and the codec configuration section (https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html#plugins-inputs-beats-codec).

Since this does not only affect the beats input, but other Logstash inputs that support multiple hosts, let's identify which ones they are and document accordingly.

acchen97:
We've always recommended demuxing with multiline on the edge with Beats, so a big +1 on documenting this better. Things break when there's load balancing, but this should be a general recommendation regardless of whether LBing is involved.

@dedemorton I've added this to our [list](https://github.com/elastic/logstash/issues/6850). As this is across Beats and LS, do you mind helping with it?

Regarding other inputs like TCP and UDP, users do place a LB to distribute load across a fleet of LS nodes, but at that point it's really the job of the user to understand their data and demux it before shipping. I'm not sure if that's something we need to document in the plugin docs, but perhaps a general recommendation that lives in the deploying and scaling guide. /cc @jordansissel

jordansissel:
I propose we instead have the beats input reject being configured with
multi line codec. We can have a failure message inform users to use multi
line in filebeat instead.

dedemorton:
+1 to Jordan's suggestion. Although we should mention in the docs that LS mutliline should not be used with Filebeat, I worry that users won't see all the juicy tidbits like this that we sprinkle through the docs.

jordansissel:
+1 also including some details in the beats input docs as well.

acchen97:
[...]

PR #202, opened 2017-04-21
Warn users that they can't use multiline codec with the beats input plugin

dedemorton:
This PR adds a statement to the Beats input plugin to warn users that they can't use the multiline codec. This PR addresses a subset of the issues tracked in logstash-plugins/logstash-input-beats#199

A couple of notes about these changes...just so you understand some of the limitations given how the doc is currently structured (tho changes are a-coming):

- I can't add this statement to the section about the codec option because the description of codec is inherited and appears in all input plugins. As a compromise I've added it to the description. We might be able to add some stuff like this by using conditional coding after we've done our big migration to move the asciidoc source out of the ruby code...but not yet.
- I haven't linked to a specific topic about configuration in the Beats guide because I'd have to link to the corresponding topic in every guide. 

@ppf2 @acchen97 Here's a statement for the beats input plugin section. Note that this content probably won't show up until 5.5 because we've already missed the freeze date for 5.4. But at least the mutl


lib/logstash/inputs/beats.rb
```diff
@@ -37,6 +37,12 @@
 # a setting for the <<plugins-inputs-beats-type,`type`>> config option in
 # Logstash, it is ignored.
 #
+# IMPORTANT: If you are shipping events that span multiple lines, you need to
+# use the configuration options available in Beats to handle multiline events
```
acchen97: If I understand correctly, we can explicitly say "Filebeat" here since it's the only official Beat with multiline support.

dedemorton: that's correct. I'll change it.

dedemorton:
@acchen97 I pushed the change you suggested. I just need a LGTM and/or approval so I can merge this. Thanks!

acchen97:
LGTM! Thanks @dedemorton.

PR #203, opened 2017-05-16
Making the plugin build system more robust and test with multiple branch and jruby version.

ph:
Allow to build the logstash with the following matrix:

- Release Gems + JRuby 1.7.25
- Unreleased gems on 5.x + JRuby-1.7.25
- Logstash Master + JRuby-1.7.25
- Specific JRuby 9k Logstash feature branch + JRuby-1.7.25

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,20 @@
 sudo: false
 language: ruby
 cache: bundler
+env:
 rvm:
-- jruby-1.7.25
-before_script:
-- bundle exec rake test:integration:setup
-- bundle exec rake vendor
-script:
-- ./gradlew test
-- bundle exec rspec spec
-- bundle exec rspec spec --tag integration
+  - jruby-1.7.25
+matrix:
+  include:
+    - rvm: jruby-1.7.25
+      env: LOGSTASH_BRANCH=master
+    - rvm: jruby-1.7.25
+      env: LOGSTASH_BRANCH=5.x
+    - rvm: jruby-9.1.9.0
+      env: LOGSTASH_BRANCH=feature/9000
+  allow_failures:
+    - rvm: jruby-9.1.9.0
+  fast_finish: true
+install: true
+script: ci/build.sh
 jdk: oraclejdk8
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+########################################################
+#
+# AUTOMATICALLY GENERATED! DO NOT EDIT
+#
+# version: 1
+########################################################
+set -e
+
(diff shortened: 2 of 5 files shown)


ci/build.sh
```diff
@@ -0,0 +1,28 @@
...
+    export LOGSTASH_SOURCE=1
+fi
+
+echo "Running install in: `pwd`"
+bundle install
+bundle exec rake vendor
+./gradlew test
+bundle exec rspec spec
+bundle exec rake test:integration:setup
+bundle exec rspec spec --tag integration
```
ph: We could split that in two, `setup / run` script, opinions?

andrewvc: I think we should split it and consider the setup script to be generated so that we can do mass updates.

Alternatively, maybe the setup script should be in a central location on github? Maybe we could just `curl` that script to prevent the need for mass updates?

andrewvc: So, maybe we just have one script that's checked into each repo which is:

```
curl http://github.com/logstash-plugins/ci/tree/master/setup.sh > setup.sh
./setup.sh
if [[ -f "run.sh" ]];
  ./run.sh
else
  bundle exec rspec spec
  bundle exec rspec spec --tag integration
fi
```

ph: I think it's easier to have 2 different files in the same repo, this will allow us to iterate and not having to synchronize another repository.

We should also easy mass update.

.travis.yml
```diff
@@ -2,13 +2,20 @@
...
+matrix:
+  include:
+    - rvm: jruby-1.7.25
+      env: LOGSTASH_BRANCH=master
+    - rvm: jruby-1.7.25
+      env: LOGSTASH_BRANCH=5.x
+    - rvm: jruby-9k
+      env: LOGSTASH_BRANCH=feature/jruby-9k
+  allow_failures:
+    - rvm: jruby-9k
```
ph: This seems to fetch jruby 9.0.0.0.pre1... I will explicitely use 9.1.9.0.

ci/build.sh
ph: I had a problem yesterday, expressing why I didn't like the ci repository to hold the script.   I think this might lead us to a situation similar to devutils. I am OK to split the logic into multiples file to allow easier edits by other developer and also make the mass update a breeze.

ph:
@andrewvc Can you do a final review?

ph:
Closing this since it will be part of a mass update.

ph:
Actually I will merge this PR, I want to do some more testing on travis.

ph:
This change was done in a mass update closing.

PR #206, opened 2017-05-29
Breaking Change: No longer support multiline codec with beats input

### LINKED ISSUE #201: Reject configuration with 'multiline' codec
related: https://github.com/logstash-plugins/logstash-input-beats/issues/199

Filebeat has multiline support, and so does Logstash. This confuses users with both choice and behavior.

Behaviors that can go wrong if you use filebeat to logstash with logstash beats input using multiline codec:

For example, If the user configures Logstash to do multiline assembly, and filebeat is not, then it is possible for a single stream (a single file, for example) to be spread across multiple Logstash instances, making it impossible for a single Logstash to reassemble.

The spread, above, can happen in at least two scenarios:
* filebeat configured without multiline and with load balancing that it spreads events across different Logstash nodes
* filebeat configured without multiline and without load balancing, a multiline event will still be multiple events within a stream, and that can be split across multiple batches to Logstash, and a network interruption will disrupt the continuity of that stream (again, only without multiline on filebeat)

For this reason, we should configure Logstash to reject the `multiline` codec with an actionable error to the user indicating that the correct way to use multiline with beats is to configure filebeat to do the multiline assembly.

### LINKED ISSUE #210: Build is Broken, Jar Dependencies are not Bundled Starting with 3.1.16
See this:

screen shot 2017-06-03 at 09 21 55

Apparently, the build stopped bundling jar dependencies breaking Core ITs with missing dependencies like here: https://travis-ci.org/elastic/logstash/jobs/238998526#L1426

```sh
[2017-06-03T04:21:41,504][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"LogStash::PluginLoadingError", :message=>"Couldn't find any input plugin named 'beats'. Are you sure this is correct? Trying to load the beats input plugin resulted in this error: Problems loading the requested plugin named beats of type input. Error: RuntimeError \n\n\tyou might need to reinstall the gem which depends on the missing jar or in case there is Jars.lock then resolve the jars with `lock_jars` command\n\nno such file to load -- io/netty/netty-all/4.1.3.Final/netty-all-4.1.3.Final (LoadError)"}
```

jakelandis:
When beats (in particular Filebeat) is not configured for multiline, incorrect groupings of data can result. When lines that should be grouped together are load balanced across multiple instances of Logstash, there is no means to properly reconstruct the grouping. It is also possible that lines that should be grouped together are not grouped properly on a single instance of Logstash due to the nature of beats batching. It is now required to handle multiline concerns in beats instead of Logstash. See https://www.elastic.co/guide/en/beats/filebeat/current/multiline-examples.html

Closes #201


ph:
Concerning theses changes, I think we will need to make a major bump and make sure Logstash prior to 6 is fixed to beats input 3.x. I am not a bit fan of having to maintain multiple branch in a plugin but in this case we don't have much choice :(

jakelandis:
> I think we will need to make a major bump...

Agreed, so that I understand the change correctly:

Bump this to 4.0.0
https://github.com/logstash-plugins/logstash-input-beats/blob/master/VERSION 

Log a new issue to Logstash, and on 5.x branch
https://github.com/elastic/logstash/blob/5.x/Gemfile.template#L98

pin the gem to 3.x
gem "logstash-input-beats", "~> 3"

...and don't change : https://github.com/elastic/logstash/blob/5.x/versions.yml since it is only for Logstash primary artifacts, right ?

> I think we can remove this line...

These lines, right ? _(not sure I trust PR line links)_

#This is the default field to which the specified codec will be applied.
config :target_field_for_codec, :validate => :string, :default => "message", :deprecated => "This option is now deprecated, the plugin is now compatible with Filebeat and Logstash-Forwarder"


> Can we update the plugin docs...

Will do.

ph:
> #This is the default field to which the specified codec will be applied.
config :target_field_for_codec, :validate => :string, :default => "message", :deprecated => "This option is now deprecated, the plugin is now compatible with Filebeat and Logstash-Forwarder"

Not that line, this [line](https://github.com/logstash-plugins/logstash-input-beats/blob/master/lib/logstash/inputs/beats.rb#L5)

ph:
@jakelandis All the other things in https://github.com/logstash-plugins/logstash-input-beats/pull/206#issuecomment-304891189 are perfect. 👍

jakelandis:
Will wait to merge till Logstash 5.x is pinned to 3.x of this gem.

ph:
@jakelandis make sure we squash the commits and update the changelog.md!

jakelandis:
Fixed up (not squashed) and updated changelog.

PR #211, opened 2017-06-03
Return proper exit code when ./gradlew vendor fails

### LINKED ISSUE #210: Build is Broken, Jar Dependencies are not Bundled Starting with 3.1.16

jakelandis:
Fixes #210

### DIFF
--- a/lib/tasks/build.rake
+++ b/lib/tasks/build.rake
@@ -3,7 +3,7 @@ require "jars/installer"
 require "fileutils"
 
 task :vendor do
-  system("./gradlew vendor")
+  exit(1) unless system './gradlew vendor'
   version = File.read("VERSION").strip
 end
 


ph:
Lets get that merged in master + 3.x branch.

PR #216, opened 2017-06-07
concurrent ruby 

ph:
The current version constraints on concurrent-ruby is really strict and
doesn't allow the core to use a newer version of the library, instead we
change this to use `~> 1.0`


logstash-input-beats.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "logstash-core-plugin-api", ">= 1.60", "<= 2.99"
 
   s.add_runtime_dependency "logstash-codec-plain"
-  s.add_runtime_dependency "concurrent-ruby", [ ">= 0.9.2", "<= 1.0.0" ]
+  s.add_runtime_dependency "concurrent-ruby", "~> 1.0"
```
jsvd: are we ok with not accepting 0.9.x? I'm guessing most logstash releases now use 0.9.x due to this restriction, will it stop this new version of being installed?

andrewvc: I agree, we should probably allow old logstashes to work here. Good call @jsvd

jsvd: I was just confirming, everything should still work since `[ ">= 0.9.2", "<= 1.0.0" ]` and `"~> 1.0"` have version `1.0.0` in common.

ph: Good call!

ph:
I need to backport that to the 3.x branch.

PR #217, opened 2017-06-07
Relax concurrent ruby versioning

ph:
The current version constraints on concurrent-ruby is really strict and
doesn't allow the core to use a newer version of the library, instead we
change this to use `~> 1.0`


ph:
backport of #216

PR #226, opened 2017-06-12
Update the matrix to build the plugin with master and jruby 9k

ph:
This will be done with a mass update scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,12 @@
 sudo: false
 language: ruby
 cache: bundler
-env:
+env: 
 rvm:
 - jruby-1.7.25
 matrix:
   include:
-  - rvm: jruby-9.1.9.0
+  - rvm: jruby-9.1.10.0
     env: LOGSTASH_BRANCH=master
   - rvm: jruby-1.7.25
     env: LOGSTASH_BRANCH=5.x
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,19 +2,15 @@
 sudo: false
 language: ruby
 cache: bundler
-env:
+env: 
 rvm:
-  - jruby-1.7.25
+- jruby-1.7.25
```
jakelandis: should this be jruby 9k ?

Also if make this change, do we need 3 builds ?  It seems that the default (env blank) would end up being the same as LOGSTASH_BRANCH=master.

ph: We actually still need 3 build, if you look at the job description:

screenshot 2017-06-13 08 36 23

The first job is;

- Test with JRuby 1.7.25 + released gems on rubyges, this build translate into the Logstash 5.4.1 release (current released stable).
- Test with JRuby 1.7.25 with the 5.x branch, which is the next 5.4.x release.
- The last test with master with JRuby 9k.

jakelandis: Thanks for explaining the difference.

jakelandis:
LGTM

ph:
This will be done in a mass-update.

PR #227, opened 2017-06-15
Provide a more complete title for search and seo

ph:
Instead of having just `Beats` we will now have `Beats input plugin`


ph:
This change will be done in a mass update.

dedemorton:
@ph LGTM. You put the AWE in awesome, man!

PR #252, opened 2017-07-27
mark congestion_threshold and target_field_for_codec as obsolete

jsvd:


jakelandis:
No need to bump to this plugin to 5.0, this plugin's 4.0 was already a major version bump intended for Logstash 6.x.

[Logstash 5.x is pinned](https://github.com/elastic/logstash/blob/5.6/Gemfile.template#L100) to this plugin's 3.x.

jsvd:
My reasoning for bumping the version is that people may be using 4.0.x at the moment, and upgrading to 4.1.x should not cause surprises. It's the blessing and curse of separate lifecycles between plugins and core

jakelandis:
LGTM

Didn't find any other uses of the newly obsoleted, and no pre-existing obsoleted.

PR #256, opened 2017-08-22
Fix exception not printed in debug log

tanji:
The condition for printing the exception in debug mode was most probably inverted, causing the exception to be printed in normal mode instead, which caused quite a headache. Also, the message is unclear, because that exception can happen in case of certificate issues too, so I added some clarification.


jordansissel:
LGTM

Issue #260, opened 2017-10-02
Timeout logic close connection too early


ph:
I've done some testing on the beats inputs; we have logic in place to close **idle remote connection**. I think the code logic is flawed and the timeout is triggered way too early than the configured timeout value which is by default 60 seconds.

I've seen it triggered after 10 seconds instead of the 60 seconds; this is problematic since the remote connection has to renegotiate a new connection and might have to retransmit batches of events.

ph:
I think a bad timeout could happen in the following situation:

- We receive events from beats.
- Logstash is stuck on the last events from a batch.
- the time to be stuck is close to the inactivity timeout.
- Logstash timeout before receiving anything.



**Solutions**

1. Use a higher timeout value
2. implement a grace period after a batch is successfully processed (invincibility frame)

I think I prefer solution 2.

ph:
I've created #262 instead which provide a better way to handle the client inactivity timeout.

ph:
Closed from #262

PR #261, opened 2017-10-02
Add contextual information in the beats handler

ph:
The current version of the logging statements in the beats handler isn't useful in the context of multiple beats client connected to the same input; this PR add the remote host information in the beats handler.

Most of the theses logging statements are only executed when running in debug mode, so the performance impact should be minimal when running with the normal log level.

Related to #260


ph:
This need to go in master and 4.x branch.

jordansissel:
Testing this:

> [2017-10-02T12:54:24,055][INFO ][org.logstash.beats.BeatsHandler] [ChannelHandlerContext(beats-handler, [id: 0x00f34543, L:0.0.0.0/0.0.0.0:5000 ! R:/0:0:0:0:0:0:0:1:45394])] Exception: not an SSL/TLS record: 68656c6c6f0d0a

^^ Sending garbage to a TLS connection

The details are not super great. Some information is useful (remote+local addresses), but hard to read. Also additional details are not useful (that it is a ChannelHandlerContext, the `id: 0x00f34543`, etc)

ph:
@jordansissel I can remove the ChannelHandlerContextBeatsHandler, but I think the following are useful `[id: 0x00f34543, L:0.0.0.0/0.0.0.0:5000 ! R:/0:0:0:0:0:0:0:1:45394]`, This allow to correlate with netty LoggingHandler.

jordansissel:
For humans, L and R are probably not the most informative (maybe say local and remote?)

What would an operator do with the `id:` value?

jordansissel:
(this all said, this PR is better than what we have today.)

ph:
`What would an operator do with the id: value?` I this is mostly for us, with a heap dump? But I think you are right, I will drop it.

Remote/Local I will go with that.

jordansissel:
Awesome :)

ph:
@jordansissel I've updated the format of the log, I also took the liberty to remove the `LoggingHandler`, this handler is way too low level to be useful in a production environment. It will just create noise.

jordansissel:
```
[2017-10-04T13:05:46,705][INFO ][org.logstash.beats.BeatsHandler] [local: 0.0.0.0:5000, remote: 0:0:0:0:0:0:0:1:34448] Exception: not an SSL/TLS record: 68656c6c6f0d0a
[2017-10-04T13:05:55,280][INFO ][org.logstash.beats.BeatsHandler] [local: 0:0:0:0:0:0:0:1:5000, remote: 0:0:0:0:0:0:0:1:34450] Exception: org.logstash.beats.BeatsParser$InvalidFrameProtocolException: Invalid Frame Type, received: 101
```

Side note: I would really prefer these to be structured logs, but I don't think we've implemented this in Java/Logstash yet, so I accept what we have for now.

LGTM

PR #262, opened 2017-10-02
Change the idle strategy

ph:
In the previous version of the beats input we were using the `IdleState.READER_IDLE`, which
is triggered when `no read was performed for the specified period of time`
This mean that we could trigger a timeout just after reading a batch
that take a long time processing.

We actually want to timeout when we didn't write (acks or keep alive) or read new events for X amount of time, so we need to use the `IdleState.ALL_IDLE`

This is important, because killing connection will make libbeat start
with a slow start strategy.

linked to #260


jordansissel:
This change looks super simple, but I need to spend some cycles learning more about netty before I can review. Shouldn't take much time.

jordansissel:
Ok I read up on IdleState and IdleStateHandler and I feel i understand thsi change more now :)

jordansissel:
LGTM

PR #263, opened 2017-10-03
Defining task using the << operator is now deprecated in Gradle

ph:
Using << for task is now deprecated in gradle you will have to use the
`doLast{}` syntax to define it.


jordansissel:
LGTM

Issue #269, opened 2017-11-17
Upgraded to 6.0 Logstash - Spammed with Could not retrieve remote IP address for beats input.


sc-chad:
In AWS using Amazon Linux for most instances.

My filebeat agents aren't upgraded yet they are still on 5.6.4.

Uninstalled previous logstash agent of 5.2.1 and did a clean install of 6.0.

Logstash input
`  beats {
    port => 5044
  }`

Results in hundreds of 
`[2017-11-17T00:52:42,143][WARN ][logstash.inputs.beats    ] Could not retrieve remote IP address for beats input.`

wf744:
I have the same issue, how to solve it?

sc-chad:
I reverted back to 5.2.0 again.  These last few versions feel like they completely skipped QA.

sc-chad:
If you want to stay on 6.0 you can do this:

Going to:  

```
/usr/share/logstash/vendor/bundle/jruby/2.3.0/gems/logstash-input-beats-5.0.2-java/lib/logstash/inputs/beats/message_listener.rb
```

And deleting: 
```
begin
    hash.get("@metadata").put("ip_address", ctx.channel().remoteAddress().getAddress().getHostAddress())
  rescue #should never happen, but don't allow an error here to stop beats input
    input.logger.warn("Could not retrieve remote IP address for beats input.")
end
```

Fixes the issue.

jordansissel:
Can you provide more details? I tried to reproduce this locally and was unsuccessful (I see no warnings).

My test setup with logstash and packetbeat:

* `bin/logstash -e 'input { beats { port => 5000 } }' (Logstash 6.0.0)
* `sudo ./packetbeat` (Packetbeat 6.0.0)

Packetbeat data is successfully transmitted and I see no warnings.

sc-chad:
Absolutely @jordansissel will help anyway I can.

We are using Filebeat from the official repos on Amazon Linux. (Same with Logstash)

An example of one of our prospectors.

```
---
filebeat:
  prospectors:
    - input_type: log
      paths:
        - /var/log/secure
      encoding: plain
      fields:
        environment: production
        product: template
        role: base
        log_type: syslog
      fields_under_root: true
      document_type: syslog
      scan_frequency: 10s
      harvester_buffer_size: 16384
      max_bytes: 10485760


      tail_files: false

      # Experimental: If symlinks is enabled, symlinks are opened and harvested. The harvester is openening the
[...]

PR #270, opened 2017-11-22
Handle Netty channels returning null for remoteAddress

robbavey:
Handles the case where the Netty channel returns remoteAddress
as null, without spamming the logs with warning messages for each
instance.

Also, changed the the method to retrieve the @metadata field from
the the messageData hash from get to '[]' - this was silently
failing in tests.

Mitigates #269


jordansissel:
> where the Netty channel returns remoteAddress as null

The description sounds like we are working around a bug in Netty. Why is `remoteAddress` null? More generally, how would we have a connection with no remote address?

robbavey:
@jordansissel There appear to be a couple of places in the Netty code where remoteAddress can return null - 
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L197 and 

https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractServerChannel.java

The comment here https://github.com/logstash-plugins/logstash-input-beats/issues/269#issuecomment-346412968 seems to point the finger at the return value from channel(). 

I tried repro'ing the issue locally, and on aws and alicloud, but to no avail...

jordansissel:
@robbavey weird; it returns null when an exception is thrown from the underlying implementation fetching the SocketAddress. I think this only happens when the socket is closed? I wonder how to safely resolve this -- we want to include the peer address in the events, and when we receive the event, the socket is open.

jordansissel:
@robbavey I'm not stating this well, having trouble with words today...

Users are reporting that this warning is "spamming" their log messages. This indicates the message happens *constantly* and the problem of a null socketAddress is frequent.  I would like us to understand why this happens now, and figure out a solution to resolve it.

If we solve it by debug logging but otherwise not warning, then these users' events will have no `ip_address` field and the next step for users is to notice that `ip_address` is missing, then file a bug about `ip_address` being missing. I feel that this PR just relocates the problem instead of fixing the underlying problem (socketAddress is null when it should not be null)

Maybe we need to cache the socketAddress when the socket is first connected? I don't know how these problems are happening, so it's hard to brainstorm a solution.

robbavey:
@jordansissel - you're right. Let me look into trying to find the root cause or find a better way to mitigate the symptoms when it happens, rather than just muting the problem. It's proving an interesting one to track down...

robbavey:
@jordansissel I've updated the code to populate the ip address on connection, rather than for every message.

cc @ph

robbavey:
ping @jordansissel (Pretty sure the build issue is not related to this PR)

robbavey:
Rebased after #275, #278

jordansissel:
Thanks; will review today!

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -83,8 +78,29 @@ def codec(ctx)
...
+        remote_address = ctx.channel.remoteAddress
+        # Netty allows remoteAddress to be nil, which can cause a lot of log entries - see
+        # https://github.com/logstash-plugins/logstash-input-beats/issues/269
+        if remote_address.nil?
+          input.logger.debug("Cannot retrieve remote IP address for beats input - remoteAddress is nil")
+          return nil
+        end
+        remote_address.getAddress.getHostAddress
+      rescue => e # This should not happen, but should not block the beats input
+        input.logger.warn("Could not retrieve remote IP address for beats input.", :message => e)
```
jordansissel: change `:message` to `:reason` or `:error` maybe?

PR #276, opened 2017-12-10
Backport: Rework the NioEventLoop handling in the beats input

robbavey:
We were using two differents NioEventLoop in the beats input, this was
 causing problems in the context of back pressure. When the NioEventLoop
 responsable of inserted to the queue was blocked we were still accepting
 new connection instead of refusing them. Creating a large amount of
 socket in the `CLOSE_WAIT` state.
    
`CLOSE_WAIT` state could also happen when an exception was raised,
 because we were flushing to the queue before closing the connection, now
 we close the connection first and flush to the queue. This can result in
 a blocked thread for some time but will reduce the connection.
This problem should only happen when using a buffering codec like the
 multiline codec.
    
We also fix an issue when the Main NioEventLoop was blocked, the keep alive signal were not send back to the client, making the beats handlers run into a separate executor allow the loop to send the keep alive signal.
 
Fixes #272

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -121,6 +121,9 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   # Close Idle clients after X seconds of inactivity.
   config :client_inactivity_timeout, :validate => :number, :default => 60
 
+  # Beats handler executor thread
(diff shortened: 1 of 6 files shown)


robbavey:
Abandoned in favor of #278

PR #283, opened 2018-01-03
Re-order pipeline to avoid NPE in KeepAliveHandler

### LINKED ISSUE #282: Under heavy load, the KeepAliveHandler can throw NullPointerExceptions
Currently, if Logstash is under heavy load, the KeepAliveHandler can throw a NullPointerException when determining whether the channel has started processing. 

This appears to be because the KeepAliveHandler is being invoked before the BeatsHandler has registered the processing attribute.

On an amended version of the codebase, the following stack trace can be observed:

```
[2018-01-03T16:48:28,002][nioEventLoopGroup-4-14][WARN ][org.logstash.beats.KeepAliveHandler] Unexpected error while checking whether batch is being processed: class java.lang.NullPointerException
java.lang.NullPointerException
	at org.logstash.beats.KeepAliveHandler.isProcessing(KeepAliveHandler.java:67) [logstash-input-beats-5.0.5.jar:?]
	at org.logstash.beats.KeepAliveHandler.userEventTriggered(KeepAliveHandler.java:22) [logstash-input-beats-5.0.5.jar:?]
	at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:329) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.channel.AbstractChannelHandlerContext.access$500(AbstractChannelHandlerContext.java:38) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.channel.AbstractChannelHandlerContext$6.run(AbstractChannelHandlerContext.java:320) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:403) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:463) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-all-4.1.18.Final.jar:4.1.18.Final]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_131]
```

robbavey:
Move the KeepAliveHandler after the BeatsHandler to avoid cases
under load where #isProcessing is called on the KeepAliveHandler
before the 'BATCH_PROCESSING' property is set in BeatsHandler,
which can cause a NullPointerException to be thrown

Also:
 Update #exceptionCaught to show the exception class if no message
is available, and to show the stack trace in the debug log.

Fixes #282

### DIFF
--- a/src/main/java/org/logstash/beats/BeatsHandler.java
+++ b/src/main/java/org/logstash/beats/BeatsHandler.java
@@ -82,7 +82,12 @@ public class BeatsHandler extends SimpleChannelInboundHandler<Batch> {
             messageListener.onException(ctx, cause);
         }
 
-        logger.info(format("Exception: " + cause.getMessage()));
+        String causeMessage = cause.getMessage() == null ? cause.getClass().toString() : cause.getMessage();
+
+        if (logger.isDebugEnabled()){
+            logger.debug(format("Handling exception: " + causeMessage), cause);
+        }
+        logger.info(format("Handling exception: " + causeMessage));
     }
 
     private boolean needAck(Message message) {
--- a/src/main/java/org/logstash/beats/Server.java
+++ b/src/main/java/org/logstash/beats/Server.java
@@ -131,9 +131,9 @@ public class Server {
             }
             pipeline.addLast(idleExecutorGroup, IDLESTATE_HANDLER, new IdleStateHandler(clientInactivityTimeoutSeconds, IDLESTATE_WRITER_IDLE_TIME_SECONDS , clientInactivityTimeoutSeconds));
             pipeline.addLast(BEATS_ACKER, new AckEncoder());
-            pipeline.addLast(KEEP_ALIVE_HANDLER, new KeepAliveHandler());
             pipeline.addLast(BEATS_PARSER, new BeatsParser());
             pipeline.addLast(beatsHandlerExecutorGroup, BEATS_HANDLER, new BeatsHandler(this.message));
+            pipeline.addLast(KEEP_ALIVE_HANDLER, new KeepAliveHandler());
         }
 
         @Override


robbavey:
Tests are now passing.

This PR has also been tested to ensure that there is no regression with the `CLOSE_WAIT` connection proliferation issue.

robbavey:
Ping @jakelandis, @jsvd  (ph is on vacation): there is [an SDH](https://github.com/elastic/support-dev-help/issues/3092) outstanding that is being affected by this bug, as well as [users referencing the issue](https://github.com/logstash-plugins/logstash-input-beats/issues/269).

jakelandis:
@robbavey - I thought the pipeline was always executed sequentially, so the new code makes sense...but how did it ever work ?

robbavey:
@jakelandis From reading the code, I think that the KeepAliveHandler.userEventTriggered gets invoked when the IdleStateHandler fires (after the channel has not written for 5 seconds). Under normal circumstances, the BeatsHandler will have set the attribute on the channel during this time, but not when Logstash is under load.

jakelandis:
LGTM

To the best of my understanding changing the order fixes this issue such that on new connections the `ctx.channel().attr(PROCESSING_BATCH)` will ensured to always be set prior trying to get the null, since we set in the handlerAdded method of the BeatsHandler, and read in the on the Fired event from the IdleStateHandler....which doesn't can't fire to the KeepAliveHandler to call the get since it doesn't exist in the pipeline until the very end.

Issue #294, opened 2018-02-07
Jackson Databind Vulnerability


ccannell67:
Upgrade Jackson Databind dependency in Logstash 5.6.x to fix deserialization vulnerability CVE-2017-7525. 
 It needs to be upgraded to one of these versions 2.6.7.1, 2.7.9.1, 2.8.9, 2.9.0

Reference [FasterXML/jackson-databind#1723](https://github.com/FasterXML/jackson-databind/issues/1723)

logstash-input-beats-3.1.26-java is using 2.7.5 and is vuln and fails vulnerability scans.
https://github.com/logstash-plugins/logstash-input-beats/blob/master/build.gradle#L30

Core is using 2.9.1 - https://github.com/elastic/logstash/blob/5.6/versions.yml

Similar and fixed in Elasticsearch plugins - see: https://github.com/elastic/elasticsearch/issues/27359

ccannell67:
Similar and fixed in Elasticsearch plugins - see: https://github.com/elastic/elasticsearch/issues/27359

jakelandis:
2.9.1 is now shipped as of 5.0.8

PR #296, opened 2018-02-08
#294 Update jackson dependencies to 2.9.1

ccannell67:


ccannell67:
The failed Travis tests seem to be with Travis, not the actual commits/pr. I have been testing against just the master and get intermittent/inconsistent passes and fails for the different versions.

Is this a known issue?

jordansissel:
> to fix vulnerability.

We are aware of this Jackson vulnerability. The issue in question is a vulnerability in the Jackson polymorphic deserialization feature. We reviewed the issue and determined we do not use this codepath in Logstash nor in Logstash beats input. It is not possible for an attacker to trigger the flaws in question. This issue does not affect Elasticsearch.

This said, I'm fine with upgrading the version of Jackson, but the reason for the upgrade claims vulnerability to something which Logstash does not appear vulnerable. Can you update the changelog to remove the CVE mention as well as the commit message to remove the vulnerability claim?

If you have verification of the claim that Logstash beats input plugin is vulnerable, please contact security@elastic.co and do not use public channels such as GitHub.

ccannell67:
OK, updated.

alex-dr:
This plugin is packaged in logstash (at least, it is in CentOS) and while I understand that this library is not affected by this class of vulnerabilities I appreciate your willingness to upgrade your dependent libraries. This helps us operate in environments where security controls, we might say, are stricter than good sense.

To that end, since something is failing in the CI build anyway, I was wondering if @ccannell67 would be willing to change the bump all the way to 2.9.4, which resolves a few more CVE's against the upstream package.

ccannell67:
Actually the core is using 2.9.1, https://github.com/elastic/logstash/blob/b3366d7ed6a71f0bceef69536c764cbc0eb2020a/versions.yml

That is why I used that version. For consistency.

alex-dr:
It's not a trivial patchset; 139 commits, nor am I set up to run tests.

https://github.com/FasterXML/jackson-databind/blob/master/release-notes/VERSION-2.x#L16-L107

I vote we also PR core. I've also got a PR to bump jrjackson since it also gets bundled in: https://github.com/guyboertje/jrjackson/pull/66

alex-dr:
Glad to hear someone else is in the same boat ;)

ccannell67:
May as well. After actually slowing down and looking at the versions/branches, this won't help me immediately anyway, as I am on ES/Logstash 5.6 and it uses logstash-input-beats version 3.x.
https://github.com/elastic/logstash/commit/eb110a2865b48b9a50a20191227d4f17302e48dc

jakelandis:
@ccannell67 - Thanks for the contribution! 

Let's get to Jackson `2.9.4` in two separate steps. Can you do the following and I will merge this PR:
* remove " to fix vulnerability." from the commit log, since there is no actual vulnerability given the scope of use, and git history is forever. 
* roll back this PR back to only bump to `2.9.1` 
* submit a new PR to bump to `2.9.4` 

I will then merge this PR and https://github.com/logstash-plugins/logstash-input-beats/pull/297 and push a release master and 3.x release with the `2.9.1` version. 

I would like to keep the Jackson version in sync with core, and core is dependent on https://github.com/guyboertje/jrjackson/pull/66.  Once that gets merged, we can bump to `2.9.4` across the stack. Whew...

ccannell67:
@jakelandis I have the commits and PR sorted. Thanks!

jakelandis:
Changes look good , thanks again @ccannell67 .  merging...

jakelandis:
This has been released.  To upgrade:
`bin/logstash-plugin update logstash-input-beats`

alex-dr:
https://github.com/guyboertje/jrjackson/pull/67

jrjackson is bumped to 2.9.4, so I guess it's time for 

> submit a new PR to bump to 2.9.4

Thanks for the quick response everyone!

jakelandis:
@alex-dr @ccannell67 - beats input Jackson dependency has been bumpted to `2.9.4` in 
beats input `3.1.29` and `5.0.10` , this should be the default versions in the next patch release of Logstash 5.x and 6.x.

PR #297, opened 2018-02-09
3.x - Update jackson deps to 2.9.1

ccannell67:
Update jackson deps to match logstash 5.x core.


ccannell67:
@jakelandis what is the process or general ETA that this will be pushed and be available in the rubygems.org repo?

jakelandis:
@ccannell67 - I just released this (was waiting on the testing to go green, we have a couple flakey tests that held this up abit)

`bin/logstash-plugin update logstash-input-beats`

Issue #286, opened 2018-01-24
Very High Memory use with Beats 6.1.x


original-brownbear:
Running Beats 6.1.x with default settings, namely `pipelining: 5` causes the beats input to consume a very large amount of memory.

This situation needs to be improved since the combination of `pipelining: x` with `x > 0` is prohibitively expensive when one Logstash instance is written to by multiple Beats clients and backpressure sets in.

Below graph shows a configuration of just Filebeat -> (beats-input -> stdout)

<img width="768" alt="screen shot 2018-01-24 at 08 16 08" src="https://user-images.githubusercontent.com/6490959/35339523-1cf9efb6-0121-11e8-8c01-55f24d411940.png">

The graph starts off showing just a single Filebeat writing to the LS instance.
At the `5:10` mark a second Filebeat is started way increasing the memory used by the input plugin.

The bulk of the memory use is made up by `org.logstash.beats.Message` instances referenced as shown below:

<img width="2277" alt="screen shot 2018-01-24 at 07 25 52" src="https://user-images.githubusercontent.com/6490959/35339629-5ef611ec-0121-11e8-9512-71b27789ca7b.png">

Possible issues I identified here are:

* We're running `Runtime.getRuntime().availableProcessors() * 4` worker threads in `Server`, that's way too much since we appear to be handling decoding (and its necessary buffering) in each of these.
* `org.logstash.beats.Message` get deserialized before use and hold redundant data as a result of `org.logstash.beats.Message#extractIdentityStream`
   * I would likely help **a lot** making the decoding here lazy and having each `org.logstash.beats.Message` simply reference a `ByteBuf` from an off-heap pool (bonus point for reusing that `ByteBuf` or even the whole `Message` instance)
* We could also optimize decoding of the actual message, constantly allocating `byte[]` for every read in `org.logstash.beats.BeatsParser#decode` puts quite some additional pressure on the GC

```java
                byte[] bytes = new byte[requiredBytes];
                in.readBytes(bytes);
                Message message = new Message(sequence, (Map) MAPPER.readValue(bytes, Object.class));
```

... I'll try to add a few more specific things I found to this issue later. The most important thing as far as I can tell from the profiling is to make `Message` cheaper by ideally just having it be a pointer to a chunk of memory (`ByteBuf` seems easiest, but slices of a `MappedByteBuffer` might be even better since it would allow paging out if necessary, I'd start with `ByteBuf` :)) that gets deserialized when used.

[...]

Issue #259, opened 2017-10-02
Logstash-input-beats connections still doesn’t close properly


amirvaza:
- Version: 5.0.1 / 3.1.24
- Operating System: ec2 linux
- Config File: follow
- Sample Data:
- Steps to Reproduce: every run

I have around 50 filebeats with this configuration:
<pre>
<code>
output:
  logstash:
    hosts: ["******:10200"]
shipper:
logging:
  level: warning
  to_files: true
  to_syslog: false
  files:
    path: /var/log/filebeat
    name: filebeat.log
    rotateeverybytes: 1000000
    keepfiles: 5
filebeat:
  prospectors:
    - paths:
        - ******
      encoding: utf-8
      input_type: log
      fields:
        type: "***"
        component: "***"
      fields_under_root: true
      multiline:
        pattern: ^[0-9][0-9][0-9][0-9]-[0-9][0-9]
        negate: true
        match: after
</code>
</pre>

And my logstash configuration:
<pre>
<code>
input {
 beats {
    port => 10200
  }
}

output
{
  elasticsearch {
    hosts => ["localhost"]
  }
</code>
</pre>

After a day or two the logstash service stops working since it has too many opened connection and I have to restart the service.
The output of netstat -p shows 16k lines of CLOSE_WAIT
After a search I saw in github a fix related to this issue in version 3.1.24 (and 5.0.1), both versions didn't seem to fix it for me.
It seems that the plugin still doesn't close the connections properly.
Or maybe I am doing something wrong?

My Logstash version is 5.6.2

Thanks.

nathanhruby:
I just hit this today with logstash 2.4.1 and logstash-input-beats 4.0.1 (this is the version I got doing an uninstall and reinstall of the plugin with bin/plugin).

In my case the connections were not closing because one of the output destinations was blocking and causing the pipeline to block (at least, a quick look of the code confirms that this is possible when the events are flushed in the close process)

This is probably something where a error level log event (or circuit breaker) should be emitted if the cnx hash is full of closers for a long time, or if flushes are not completing in a timely fashion.

jordansissel:
the beats input already has a circuit breaker to reject new connections when the internal queue is full. I recommend upgrading Logstash.

jordansissel:
> The output of netstat -p shows 16k lines of CLOSE_WAIT


I'm not sure why this happened. Please provide full details of your deployment (beats config, Logstash config, any source of network interference[1] between beats and logstash).

[...]

PR #299, opened 2018-02-13
Memory management and back pressure improvements

robbavey:
The main intent of this commit is to mitigate the very high memory usage detailed in #286, by ensuring that the KeepAliveHandler works when Logstash queues are blocked, and improve the memory usage by using `ByteBuf` instances where possible, and deserializing as late as possible.

Also:
  Reduces the default number of beats executor threads from
4*number_of_processors to number_of_processors
  Includes fixes for #259
 Fixes an integration test which was failing regularly on Travis

### DIFF
--- a/src/main/java/org/logstash/beats/Message.java
+++ b/src/main/java/org/logstash/beats/Message.java
@@ -4,11 +4,9 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufInputStream;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 
 import java.io.IOException;
-import java.util.HashMap;
+import java.io.InputStream;
 import java.util.Map;
 
 public class Message implements Comparable<Message> {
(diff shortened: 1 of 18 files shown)


src/main/java/org/logstash/beats/Message.java
```diff
@@ -1,27 +1,65 @@
...
 
-
-    public Map getData() {
+    /**
+     * Returns a list of key/value pairs representing the contents of the message.
+     * Note that this method is lazy if the Message was created using a {@ByteBuf}
+     * @return {@link Map} Map of key/value pairs
+     */
+    public Map getData(){
+        if (data == null && buffer != null){
```
original-brownbear: @robbavey don't we have to release this `ByteBuf` after we've deserialized its contents?

robbavey: This is a `slice`d from the Batch, which has its own separate `release` method to be called after it has been processed

original-brownbear: @robbavey ah right thanks :)

original-brownbear:
@robbavey do you have some before after data for the memory behaviour change here?

Ideally, some graph like the one I posted in that shows the change in memory use when moving from a single Beat to two Beats?

robbavey:
screen shot 2018-02-13 at 11 50 18 pm
@original-brownbear First Beats inputs was added at ~1:00, 2nd at 1:35, 3rd at ~2:20

original-brownbear:
@robbavey thanks, that does look better than before I think. I have to try this out locally since my test run was with a lot lower `Xmx` and triggered old gen GC, but it seems at least adding another Beat doesn't trigger a more than proportional increase in memory use :)

robbavey:
@original-brownbear Sorry - missed the axes off the graph - that is with `-Xmx1g`, which is the same as the graph in #286

original-brownbear:
@robbavey hehe even better then :D

src/main/java/org/logstash/beats/Message.java
```diff
@@ -30,20 +68,23 @@ public int compareTo(Message o) {
...
     public String getIdentityStream() {
+        if (identityStream == null){
+            identityStream = extractIdentityStream();
+        }
         return identityStream;
     }
 
     private String extractIdentityStream() {
-        Map beatsData = (HashMap<String, String>) this.getData().get("beat");
+        Map beatsData = (Map)this.getData().get("beat");
```
original-brownbear: @robbavey why drop the type information here? Shouldn't this be `Map<String, String>`?

src/main/java/org/logstash/beats/Server.java
```diff
@@ -34,25 +31,22 @@
 
     private final int clientInactivityTimeoutSeconds;
 
-    public Server(String host, int p) {
-        this(host, p, DEFAULT_CLIENT_TIMEOUT_SECONDS, Runtime.getRuntime().availableProcessors() * 4);
-    }
-
-    public Server(String host, int p, int timeout, int threadCount) {
+   public Server(String host, int p, int timeout, int threadCount) {
+//       ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.PARANOID);
```
original-brownbear: @robbavey I guess this should be cleaned up before merging :)

src/main/java/org/logstash/beats/Server.java
```diff
@@ -34,25 +31,22 @@
...
         this.host = host;
         port = p;
         clientInactivityTimeoutSeconds = timeout;
         beatsHeandlerThreadCount = threadCount;
         workGroup = new NioEventLoopGroup();
-    }
+   }
 
-    public void enableSSL(SslSimpleBuilder builder) {
+    void enableSSL(SslSimpleBuilder builder) {
```
original-brownbear: @robbavey any particular reason for changing the permissions here?

src/main/java/org/logstash/beats/Message.java
original-brownbear: @robbavey I think you could just `null` out `data` here though once it's not used anymore i.e. after deserialization. It's a minor win but it also makes the intention behind lazy deserialization more visible imo

original-brownbear:
@robbavey in general, this looks like a step in the right direction. There's quite a bit of noise in the PR though (changed visibilities mostly and some formatting + leftover debugging line). Can you revert that stuff? Other than that I'd merge this and try to make further progress on the memory front in subsequent PRs :)

src/main/java/org/logstash/beats/Message.java
robbavey: Will change

src/main/java/org/logstash/beats/Server.java
robbavey: 👍

src/main/java/org/logstash/beats/Server.java
robbavey: Just trying to remove unnecessarily permissive access modifiers. Will revert these changes and follow up later.

src/main/java/org/logstash/beats/Message.java
robbavey: `getData` is called more than once - doesn't it make sense to keep it around for subsequent access?

original-brownbear: @robbavey sorry my bad. I meant null out `buffer` since we don't use that more than once :)

robbavey: @original-brownbear 👍 That makes a lot more sense ;)

robbavey:
@original-brownbear New commit pushed

Thanks for the quick review, btw.

original-brownbear:
@robbavey np, LGTM if it goes green :)

Issue #326, opened 2018-06-21
Setting ssl_verify_mode to PEER has no effect and will act as FORCE_PEER


tsouza:
Setting `ssl_verify_mode` to `peer` has no effect and will act as `force_peer`

Version: `6.3.0`

Steps to reproduce:
1. In `logstash.conf`:
```ruby
input {
	beats {
		port => 5000
		ssl => true
		ssl_certificate => "/path/to/cert.pem"
		ssl_key => "/path/to/key.pem"
		ssl_certificate_authorities => "/path/to/ca.pem"
		ssl_verify_mode => "peer"		
	}
}
```
2. In `filebeat.yml`:
```yaml
filebeat.inputs:
  - type: stdin
output.logstash:
  hosts: ["127.0.0.1:5000"]
  ssl.certificate_authorities: ["/path/to/ca.pem"]
```

After typing in Filebeat's stdin, Logstash logs:
```
[2018-06-21T01:27:27,659][INFO ][org.logstash.beats.BeatsHandler] [local: 0.0.0.0:5000, remote: 127.0.0.1:53279] Handling exception: javax.net.ssl.SSLHandshakeException: error:100000c0:SSL routines:OPENSSL_internal:PEER_DID_NOT_RETURN_A_CERTIFICATE
[2018-06-21T01:27:27,660][WARN ][io.netty.channel.DefaultChannelPipeline] An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: error:100000c0:SSL routines:OPENSSL_internal:PEER_DID_NOT_RETURN_A_CERTIFICATE
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:459) ~[netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:265) ~[netty-all-4.1.18.Final.jar:4.1.18.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362) ~[netty-all-4.1.18.Final.jar:4.1.18.Final]
[...]

Issue #83, opened 2016-05-03
Put TLS connection info in message metadata


stith:
Hey Y'all!

With the new `@metadata` field it seems like Logstash could expose some potentially-useful info regarding the status of the TLS connection there, for cases where certificate-based verification of message sources might be important. For example, when `ssl_verify_mode => 'peer'`, a beat could connect with a client cert or without one, and it might be nice to know:

```json
{
  "message": "Test message",
  "@metadata": {
    "tls_client_status": "verified",
    "tls_subject_cn": "some-server-zzzz.company.com"
  }
}
```
Or,
```json
"@metadata": {
  "tls_client_status": "unverified"
}
```

erikswanson:
My use case for wanting to know the client certificate CN is forcing `host` and `beat.hostname` to match the client certificate's CN, to prevent forgery of log message origin in the event of the compromise of a host with a valid client certificate. (I could trivially accomplish my use case if the proposal above was implemented, but am stuck on being able to use beats for security logging without it. ☹️)

I think it also might work if there was _just_ a `tls_subject_cn` field that was empty or absent in the case of no client certificate, instead of including a seemingly-redundant status field.

directionless:
I was just looking for this feature! I'd love to see this. I'm looking at using logstash and filebeat to gather data from various hosts. I think the only non-forgeable piece of information is the TLS CN. It's a great identifier, and I've to have it in the metadata

directionless:
I started playing around. https://github.com/logstash-plugins/logstash-input-beats/compare/master...directionless:seph/tls-metadata seems to work, though I'm not sure how to setup a good test suite. Or whether there are performance impacts.

ansoni:
+1 Once we have this, I can use the certificate CN as a way to lookup metadata within logstash and enrich the input message.  This would be pretty awesome.

PR #327, opened 2018-06-21
Add support to storing TLS client authentication in event's metadata

tsouza:
This PR is a proposal for adding the feature described here https://github.com/logstash-plugins/logstash-input-beats/issues/83

Unfortunately I could not fully test it due to https://github.com/logstash-plugins/logstash-input-beats/issues/326

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -89,7 +89,7 @@ input plugins.
 &nbsp;
 
 [id="plugins-{type}s-{plugin}-cipher_suites"]
-===== `cipher_suites` 
+===== `cipher_suites`
 
   * Value type is <<array,array>>
   * Default value is `java.lang.String[TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256, TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256]@459cfcca`
@@ -97,7 +97,7 @@ input plugins.
 The list of ciphers suite to use, listed by priorities.
 
 [id="plugins-{type}s-{plugin}-client_inactivity_timeout"]
-===== `client_inactivity_timeout` 
+===== `client_inactivity_timeout`
 
   * Value type is <<number,number>>
   * Default value is `60`
@@ -105,7 +105,7 @@ The list of ciphers suite to use, listed by priorities.
 Close Idle clients after X seconds of inactivity.
 
 [id="plugins-{type}s-{plugin}-host"]
-===== `host` 
+===== `host`
 
   * Value type is <<string,string>>
   * Default value is `"0.0.0.0"`
@@ -113,7 +113,7 @@ Close Idle clients after X seconds of inactivity.
 The IP address to listen on.
 
 [id="plugins-{type}s-{plugin}-include_codec_tag"]
-===== `include_codec_tag` 
+===== `include_codec_tag`
 
   * Value type is <<boolean,boolean>>
   * Default value is `true`
@@ -121,7 +121,7 @@ The IP address to listen on.
 
 
 [id="plugins-{type}s-{plugin}-port"]
-===== `port` 
+===== `port`
 
   * This is a required setting.
   * Value type is <<number,number>>
@@ -130,7 +130,7 @@ The IP address to listen on.
 The port to listen on.
 
 [id="plugins-{type}s-{plugin}-ssl"]
-===== `ssl` 
+===== `ssl`
 
   * Value type is <<boolean,boolean>>
   * Default value is `false`
@@ -140,7 +140,7 @@ enable encryption by setting `ssl` to true and configuring
 the `ssl_certificate` and `ssl_key` options.
 
 [id="plugins-{type}s-{plugin}-ssl_certificate"]
-===== `ssl_certificate` 
+===== `ssl_certificate`
 
   * Value type is <<path,path>>
   * There is no default value for this setting.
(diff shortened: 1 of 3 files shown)


lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -119,6 +122,35 @@ def flush_buffer(ctx)
...
+        rescue SSLPeerUnverifiedException => e
+          if input.logger.debug?
+            input.logger.warn("Failed to get peer certificates.", :exception => e.to_s)
+          end
+        end
+
+        if tls_verified
+          hash['@metadata']['tls_peer'] = {
+            :status     => "verified",
+            :protocol   => tls_session.getProtocol(),
```
jakelandis: would this always be `https` ?

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -119,6 +122,35 @@ def flush_buffer(ctx)
...
+            :principal  => tls_session.getPeerPrincipal().getName()
```
jakelandis: Is this the CN of the peer cert ?

lib/logstash/inputs/beats/message_listener.rb
tsouza: If you mean `protocol` then no. This is the actual TLS protocol version being used, like `TLSv1.2`

lib/logstash/inputs/beats/message_listener.rb
tsouza: Not only `CN`. It should be the complete subject `DN`, that also includes a `CN`. This is an arbitrary that client may inform to identify itself

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -119,6 +122,35 @@ def flush_buffer(ctx)
...
+    def extract_tls_peer(hash)
+      if @input.client_authentication_metadata?
+        tls_session = ctx.channel().pipeline().get("ssl-handler").engine().getSession()
+        tls_verified = false
+
+        # throws SSLPeerUnverifiedException if unverified
+        begin
+          tls_session.getPeerCertificates()
+          tls_verified = true
+        rescue SSLPeerUnverifiedException => e
```
jakelandis: In JRuby exceptions are very expensive, is there an alternative way to check if a peer certificate was presented ?

lib/logstash/inputs/beats/message_listener.rb
jakelandis: doh! wrong plugin, this plugin is not even http based :) 

Is there an option for the cipher chosen too ?

lib/logstash/inputs/beats/message_listener.rb
tsouza: In theory, yes. Netty should raise a specific `HandshakeCompletionEvent` event in https://github.com/logstash-plugins/logstash-input-beats/blob/master/src/main/java/org/logstash/beats/ConnectionHandler.java#L64 which the certificate data can be extracted in the very same way.

The problem is that I don't know how to assign the data to the event as there are no obvious references to the event object at that method.

lib/logstash/inputs/beats.rb
```diff
@@ -93,6 +93,10 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   # This option needs to be used with `ssl_certificate_authorities` and a defined list of CAs.
   config :ssl_verify_mode, :validate => ["none", "peer", "force_peer"], :default => "none"
 
+  # Enables storing client certificate information in event's metadata. You need 
+  # to configure the `ssl_verify_mode` to `peer` or `force_peer` to enable this.
```
jakelandis: Can we enforce this in the register method with a `raise LogStash::ConfigurationError ...`

tsouza: Sure!

lib/logstash/inputs/beats/message_listener.rb
tsouza: I would say that this is an edge case anyway. It means that the client certificate metadata is being asked while allowing to _not_ use a client certificate. I would say that this would be used more for testing then anything else.

BTW `ssl_verify_mode => "peer"` does not even works and this was never reported. See https://github.com/logstash-plugins/logstash-input-beats/issues/326

tsouza: Maybe there could be an optimization of checking `force_peer` and then never call this `getPeerCertificates()` method (I am not really sure what happens under the covers)

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -119,6 +122,35 @@ def flush_buffer(ctx)
...
+          if input.logger.debug?
+            input.logger.warn("Failed to get peer certificates.", :exception => e.to_s)
```
jakelandis: Can you move this to just a debug message, and mention that for 'peer' mode this may be expected. Something like this: (also shouldn't need to  .to_s )
```
input.logger.debug("Failed to get peer certificates. This is normal for 'peer' verification", :exception => e)
```

jakelandis:
@tsouza - changes look good. minor changes requested. 

Also, can you please update the [docs](https://github.com/logstash-plugins/logstash-input-beats/blob/master/docs/index.asciidoc) with the new option ?

I will make a note to tests and merge soon.

lib/logstash/inputs/beats/message_listener.rb
tsouza: Yes, there is! I will add it.

docs/index.asciidoc
```diff
@@ -200,6 +201,16 @@ If the client doesn't provide a certificate, the connection will be closed.
...
 
+[id="plugins-{type}s-{plugin}-ssl_peer_metadata"]
+===== `ssl_peer_metadata` 
+
+  * Value type is <<boolean,boolean>>
+  * Default value is `false`
+
+Enables storing client certificate information in event's metadata.
+
+This option needs to be used with `ssl_verify_mode` set to `peer` or `force_peer`.
```
jakelandis: nitpick: 

> This option needs to be used with `ssl_verify_mode` set to `peer` or `force_peer`.

to

> This option is only valid when `ssl_verify_mode` is set to `peer` or `force_peer`.

jakelandis:
@tsouza - Code changes look good. Tested it manually and it all works as documented. 

one very minor nitpick on the wording of the doc. 

Please squash/fixup the commits and I will get it merged and released.

jakelandis:
ooops wrong button! (re-opening)

jakelandis:
This has been release in version 5.1.0.
To upgrade run the following command:
```
bin/logstash-plugin update logstash-input-beats
```

PR #329, opened 2018-06-21
Increased use of log4j2 readability.

fbacchella:
Added a log4j2.xml for tests runs.

Added a log4j2.xml for tests runs.

The plugin does not actually uses the log4j2 full power, and that make the code less readable.


karmi:
Hi @fbacchella, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-beats/pull/329.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @fbacchella, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-beats/pull/329.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

fbacchella:
CLA signed.

fbacchella:
It needs PR #330 to work.

src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
robbavey: Can we split this change out into a separate PR? The validity of adding debug logging to test cases is a different concern to a code cleanup

PR #330, opened 2018-06-21
channel().localAddress() not always return an InetSocketAddress

fbacchella:
channel().localAddress() not always return an InetSocketAddress, for example during unit tests.

So check they are the right type before type cast.

It must be merged before PR #329.


fbacchella:
Travis failed only for jruby 1.7.27. All others succeeded. Is that something related to this PR or more generic .

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -114,20 +116,22 @@ private void writeAck(ChannelHandlerContext ctx, byte protocol, int sequence) {
...
-            localhost = local.getAddress().getHostAddress() + ":" + local.getPort();
-        } else{
+        SocketAddress local = context.channel().localAddress();
+        SocketAddress remote = context.channel().remoteAddress();
+
+        String localhost ;
+        if (local instanceof InetSocketAddress) {
+            InetSocketAddress inetlocal = (InetSocketAddress)local;
+            localhost = inetlocal.getAddress().getHostAddress() + ":" + inetlocal.getPort();
+        } else {
```
robbavey: Could you refactor this to remove the code duplication?

fbacchella:
$ rvm use jruby-9.1.13.0 --install --binary --fuzzy
curl: (22) The requested URL returned error: 404 Not Found

The travis failure is not mine.

PR #340, opened 2018-07-09
Make adding the hostname to the event optional

robbavey:
Add option to make the population of the 'host' field in the event with
the value from `beat.hostname` optional. This field is intentionally added
as a deprecated field, as the this field will be removed in future versions
of the beats input, when the ability to populate `host` field in this way
will be removed altogether.


lib/logstash/inputs/beats.rb
```diff
@@ -82,6 +82,10 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   # 
   config :ssl_certificate_authorities, :validate => :array, :default => []
 
+  # Flag to determine whether to add host information (provided by the beat in the 'hostname' field) to the event
+  config :add_hostname, :validate => :boolean, :default => true, :deprecated => 'Host field will not be automatically populated by future version of the Beats input'
```
danhermann: Doesn't the default value of `true` conflict with the documented default value?

robbavey: Gah, you're right. Good catch, will change

robbavey:
@karenzone Would you mind checking the doc changes?

docs/index.asciidoc
```diff
@@ -112,6 +113,18 @@ input plugins.
 
 &nbsp;
 
+[id="plugins-{type}s-{plugin}-add_hostname"]
+===== `add_hostname`
+
+  added[5.2.0, Field was added to allow users to not automatically add the `host` field in events.]
```
karenzone: What about "Field was added to allow users to control whether or not the `host` field is automatically added to events."

docs/index.asciidoc
```diff
@@ -112,6 +113,18 @@ input plugins.
...
+  deprecated[5.2.0, In future versions of this plugin, this setting will be removed, and the 'hosts' field will not be added to events.]
```
karenzone: Can we say `deprecated[5.2.1` if the field will be removed for *future* releases? 
To summarize suggestion, "Added in 5.2.0.  Deprecated in 5.2.1"

docs/index.asciidoc
robbavey: +1 - I like that

docs/index.asciidoc
robbavey: The intent is to deprecate it immediately - the setting is a stop gap until we make the breaking change that will remove the setting.

PR #334, opened 2018-06-25
Upgrade to Gradle 4.8.1

praseodym:
Refer to https://gradle.org/releases/ for release notes.

PR #341, opened 2018-07-09
Version bump and changelog entries

robbavey:
Add version bump and changelog entry for gradle upgrade added in #334


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.1.2
+ - Changed project to use gradle version 4.8.1. [#334](https://github.com/logstash-plugins/logstash-input-beats/pull/334)
```
yaauie: While this is _accurate_, it's not particularly _helpful_.

When I go to a changelog, it's typically because I'm attempting to determine if it is either safe or helpful to upgrade the thing in question, and I usually don't have the mental context to know that version X of some dependency is "better" or "worse" than whatever it was using before.

 - Why did we change it? Is it consuming an upstream fix (and for what, specifically)?
 - Is this a plugin-developer-only change, or does it affect users who install the plugin?

robbavey:
@yaauie Thanks!

(And thanks for being thorough, and looking out for our users)

PR #342, opened 2018-07-11
Fix sending of acks when sequence number of batch does not start with 1

robbavey:
Acknowledgements back to the sender were only being sent when the sequence
number of the message being handled was equal to the size of the batch.
This was causing a problem with the lumberjack output, as the underlying
jls-lumberjack library does not reset the sequence number to 1 for each
batch. This was causing a signficant slowdown in the lumberjack output to
beats input pipeline.


colinsurprenant:
Overall LGTM - I would suggest adding test coverage for the `highestSequence` and `get HighestSequence` logic.

colinsurprenant:
LGTM<sup>2</sup>

PR #345, opened 2018-07-16
Version bump to 3.1.32

robbavey:
And update changelog


colinsurprenant:
LGTM

colinsurprenant:
Just to make sure - there is not potential BWC  issue with #342 right?

robbavey:
@colinsurprenant No - it fixes compatibility with the lumberjack output, which was previously broken.

I've tested against filebeat and logstash with a configuration setup with the lumberjack output.

Thanks for the LGTM

PR #347, opened 2018-07-25
Remove extra space to fix broken doc versioned plugin reference build

dedemorton:
The extra space in the link was causing problems with the versioned plugin reference build because the cleanup part of the script was unable to make the necessary replacement: https://github.com/elastic/docs-tools/blob/master/versioned_plugins.rb#L224


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 5.1.6
+ - Docs: Fix broken link by removing extra space.
```
robbavey: Would you mind changing this to use the format described in [the logstash plugin contribution guidelines](https://github.com/elastic/logstash/blob/master/CONTRIBUTING.md#what-is-the-format-of-the-changelog-entries) for writing changelogs?

dedemorton:
@robbavey Is this what you had in mind? https://github.com/logstash-plugins/logstash-input-beats/pull/347/commits/29ff0bca62a16f1e2ddf1a9bc9b83f60e86574b5

PR #350, opened 2018-10-27
When creating the private BeatsInitializer, some attributes from the wrapping server are copied, some none.

fbacchella:
So changing some attributes in the server will change the internal running state, some will not.
This patch choose to copy everything.

### DIFF
--- a/src/main/java/org/logstash/beats/Server.java
+++ b/src/main/java/org/logstash/beats/Server.java
@@ -110,15 +110,15 @@ public class Server {
 
         private final EventExecutorGroup idleExecutorGroup;
         private final EventExecutorGroup beatsHandlerExecutorGroup;
-        private final IMessageListener message;
-        private int clientInactivityTimeoutSeconds;
+        private final IMessageListener localMessageListener;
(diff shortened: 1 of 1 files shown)


src/main/java/org/logstash/beats/Server.java
```diff
@@ -110,15 +110,15 @@ public boolean isSslEnable() {
...
-
-        private boolean enableSSL = false;
-
-        public BeatsInitializer(Boolean secure, IMessageListener messageListener, int clientInactivityTimeoutSeconds, int beatsHandlerThread) {
-            enableSSL = secure;
-            this.message = messageListener;
-            this.clientInactivityTimeoutSeconds = clientInactivityTimeoutSeconds;
+        private final IMessageListener localMessageListener;
+        private final int localClientInactivityTimeoutSeconds;
+        private boolean localEnableSSL;
```
robbavey: This can be `final` as it is initialized in the constructor

fbacchella: Indeed, I missed this one.

PR #359, opened 2019-02-15
Improve consistency of messaging around `ssl_verify_mode`

robbavey:
Ensure that log messages use the correct settings and values for `ssl_verify_mode`. Use downcased version of `force_peer` and `peer` in ruby code and logs consistently rather than mixing cases.


jsvd:
while we're here, maybe change things [here](https://github.com/logstash-plugins/logstash-input-beats/blob/21ba198ffb900f363c955a4e260a215c8b3f7b3e/lib/logstash/inputs/beats.rb#L180-L184) too (where we force the upcase on the string before comparing)

PR #378, opened 2020-01-20
Downgrade netty due to IdleStateHandler issue

jsvd:
It appears sending lumberjack payloads to a beats input running netty 4.1.35 or newer causes the acknowledge to take a long time to be written back to the client, often causing i/o timeouts. 

One can find more information in https://github.com/netty/netty/issues/8912 and https://github.com/elastic/logstash/issues/11517.

While further investigation is done, this PR downgrades netty to .34


jsvd:
a script to replicate this slow acknowledgement can be found in https://gist.github.com/jsvd/4315e10241049ff0e55898a2f3797542

PR #384, opened 2020-02-03
drop dependency on tcnative+boringssl

jsvd:


jsvd:
replaced by https://github.com/logstash-plugins/logstash-input-beats/pull/393

PR #388, opened 2020-03-06
Remove `java_import`s to avoid breaking tcp input

robbavey:
Since the release of `6.0.8`, when pipelines (or across multiple pipelines), when a tcp input and beats input are defined, both using TLS, the TCP input uses org.logstash.netty.SslContextBuilder defined here instead of the desired SslContextBuilder from the Netty library, which causes TCP pipelines to fail with `NoMethodError`.
    
This commit removes the (unnecessary) `java_imports` from the `beats.rb` to avoid this.
    
This commit also pins devutils to `~>1.3` to fix the build.


kares:
interesting, looking through both the plugins (trying to understand what's going on) 
this seems as a class-loading issue rather than a simple name collision, isn't that right?

robbavey:
@kares The error that is produced as a result of this issue is:

```
exception=>#<NoMethodError: undefined method `forServer' for Java::OrgLogstashNetty::SslContextBuilder:Class>
```

(note: `OrgLogstashNetty` rather than `io.netty.handler.ssl.SslContextBuilder`)

which is emitted from https://github.com/logstash-plugins/logstash-input-tcp/blob/master/lib/logstash/inputs/tcp/compat_ssl_options.rb#L110

kares:
oh thanks - that makes sense. 
very weird - might try a simple reproducer and raise upstream w JRuby.

PR #393, opened 2020-05-11
Update/remove dependencies

robbavey:
This commit updates the following dependencies:

Jackson-databind to `2.9.10.4`
Netty to `4.1.49.final`

Additionally, this commit removes the dependency on `tcnative` +
`boringssl` based on the work done by @jsvd in #384, which enables the update of netty dependencies while still allowing logstash->logstash communication via the lumberjack
output plugin. (`boringssl` removed compatibility with `CBC` ciphers, which
are still a prerequisite for lumberjack output plugin support at this time)


robbavey:
In addition to the integration tests, this commit was tested against [this](https://github.com/logstash-plugins/logstash-input-beats/pull/378#issuecomment-576350190) script created by @jsvd to exercise [this issue](https://github.com/elastic/logstash/issues/11517) where the IdleStateHandler was holding back acknowledgements. 

Running this script to completion:

Beats Input `6.0.5`:
```
real	1m3.108s
user	0m10.231s
sys	0m0.539s
```

`6.0.9`:
```
real	0m3.223s
user	0m10.363s
sys	0m0.485s
```

This commit:
```
real	0m3.231s
user	0m10.295s
sys	0m0.585s
```

robbavey:
@jsvd - one thing to note, this will break on versions of Java 8 without the JCE Unlimited Strength Jurisdiction Policy installed. From `Java 1.8.0_162` onwards (released January 2018), the policy was installed by default, but prior to this release, these ciphers are not all present, leading to a plugin failure. 

I can update the plugin to be more lenient, and warn on missing ciphers, and only fail when no ciphers are available. WDYT?

jsvd:
I think it's fine that we detect if jce isn't available (maybe through the maximum cipher lenght which should be 128 w/o jce), we exclude the aes 256 ciphers:
```
          "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
          "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
          "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384",
          "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384",
```

robbavey:
@jsvd I've made the changes and updated the changelog and bumped the version - do you think the version should be a minor or patch update?

jsvd:
I'd vote on patch as a it's a fix and doesn't introduce any new user facing capability

robbavey:
Done

PR #394, opened 2020-05-12
detect if ssl certificate or key can't be read

### LINKED ISSUE #197: having issues with SSL options - what formats are allowed?
```
9:37:08.516 [nioEventLoopGroup-4-1] ERROR logstash.inputs.beats - Looks like you either have an invalid key or your private key was not in PKCS8 format. {:exception=>java.lang.IllegalArgumentException: File does not contain valid certificates: /etc/pki/tls/certs/{REDACTED}.der.crt}
```
why is it saying my key is invalid and then saying my cert file doesn't contain valid certificates? what forms are required for options (DER or PEM?)

jsvd:
In the scenario where either the ssl certificate or key files can't be read
by the user running Logstash, e.g. due to permissions, an error is generated
by io.netty.handler.ssl.SslContextBuilder.forServer communicating that these
files don't contain valid content:

"java.lang.IllegalArgumentException: File does not contain valid certificates"
or
"java.lang.IllegalArgumentException: File does not contain valid private key"

This is often misleading, causing users to question if their certificates are
created incorrectly. So this commit checks first if these files can be read,
and throws an exception dedicated to this error.

Before:

> [2020-05-11T14:53:24,379][ERROR][logstash.javapipeline    ][main] Pipeline aborted due to error {:pipeline_id=>"main", :exception=>java.lang.IllegalArgumentException: File does not contain valid private key: /tmp/logstash-7.6.2/Server-Root.key.pkcs8

After:

> [2020-05-12T09:51:49,420][ERROR][logstash.javapipeline    ][main] Pipeline aborted due to error {:pipeline_id=>"main", :exception=>#<LogStash::ConfigurationError: Private key file cannot be read. Please confirm the user running Logstash has permissions to read: /tmp/logstash-7.6.2/Server-Root.key.pkcs8>

closes #197


andsel:
LGTM

Issue #400, opened 2020-11-27
key fatal error not printed with default info level


kares:
running beats with TLS configured:
```
  beats {
    id => "beats-tls"
    port => 5044

    ssl => true
    ssl_key => "/cert-bundle/logstash.key"
    ssl_certificate => "/cert-bundle/logstash.crt"
  }
}
```
and give the key is not in the expected format LS will only log this much:
```
[2020-11-27T14:54:15,359][INFO ][logstash.inputs.beats    ][main] Beats inputs: Starting input listener {:address=>"0.0.0.0:5044"}
[2020-11-27T14:54:15,639][ERROR][logstash.agent           ] Failed to execute action {:id=>:main, :action_type=>LogStash::ConvergeResult::FailedAction, :message=>"Could not execute action: PipelineAction::Create<main>, action_result: false", :backtrace=>nil}
[2020-11-27T14:54:15,810][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2020-11-27T14:54:20,876][INFO ][logstash.runner          ] Logstash shut down.
[2020-11-27T14:54:20,915][ERROR][org.logstash.Logstash    ] java.lang.IllegalStateException: Logstash stopped processing because of an error: (SystemExit) exit
```
---
only with `--log.level debug` the key issue (`java.lang.IllegalArgumentException: File does not contain valid private key`) is visible:
```
[...]

PR #402, opened 2020-11-27
Fix: log error when SSL context building fails

kares:
There's several changes to make sure plugin provides better user-experience when it comes to logging.
The main motivator for the change set are key issues such as `java.lang.IllegalArgumentException: File does not contain valid private key` which are only logged upstream at a debug level (see #400).

On LS 7.8 / 7.9 the output when key has invalid format (or a ssl_xxx related configuration issue looks like):
(one needs to use `--log.level debug` to find out what the issue was)

```
[2020-11-27T15:40:32,191][INFO ][logstash.javapipeline    ][main] Pipeline Java execution initialization time {"seconds"=>0.52}
[2020-11-27T15:40:32,211][INFO ][logstash.inputs.beats    ][main] Starting input listener {:address=>"0.0.0.0:5044"}
[2020-11-27T15:40:32,308][ERROR][logstash.inputs.beats    ][main] SSL configuration invalid {:message=>"File does not contain valid private key: /home/kares/workspace/work/elastic/plugins/logstash-input-beats/cert-bundle/logstash.key"}
[2020-11-27T15:40:32,485][ERROR][logstash.agent           ] Failed to execute action {:id=>:main, :action_type=>LogStash::ConvergeResult::FailedAction, :message=>"Could not execute action: PipelineAction::Create<main>, action_result: false", :backtrace=>nil}
[2020-11-27T15:40:32,652][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2020-11-27T15:40:37,717][INFO ][logstash.runner          ] Logstash shut down.
[2020-11-27T15:40:37,730][ERROR][org.logstash.Logstash    ] java.lang.IllegalStateException: Logstash stopped processing because of an error: (SystemExit) exit
```

Thus to improve the logger experience there's 3 set of changes:

- log when plugin's SSL context builder fails (user can easily spot key format errors)
- make sure we log configuration errors before raising them (due https://github.com/elastic/logstash/issues/12484)
- on SSL hand-shake failures extract the real-cause as an attempt to log a better info message e.g.

`[2020-11-30T12:53:38,997][INFO ][org.logstash.beats.BeatsHandler][main][beats-tls] [local: 0.0.0.0:5044, remote: 192.168.122.59:57466] Handling exception: io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: General SSLEngine problem (caused by: java.security.cert.CertPathValidatorException: Path does not chain with any of the trust anchors)`

resolves https://github.com/logstash-plugins/logstash-input-beats/issues/400


lib/logstash/inputs/beats.rb
```diff
@@ -131,44 +131,40 @@ def register
...
 
-    if !@ssl
-      @logger.warn("Beats input: SSL Certificate will not be used") unless @ssl_certificate.nil?
-      @logger.warn("Beats input: SSL Key will not be used") unless @ssl_key.nil?
-    elsif !ssl_configured?
-      raise LogStash::ConfigurationError, "Certificate or Certificate Key not configured"
-    end
+    if @ssl
+      if @ssl_key.nil? || @ssl_key.empty?
+        configuration_error "Missing required setting ssl_key => ... (due ssl => true)"
```
robbavey, line 136: Maybe could be reworded a little?

```suggestion
        configuration_error "ssl_key => is a required setting when ssl=>true is configured"
```

lib/logstash/inputs/beats.rb
```diff
@@ -131,44 +131,40 @@ def register
...
+      end
+      if @ssl_certificate.nil? || @ssl_certificate.empty?
+        configuration_error "Missing required setting ssl_certificate => ... (due ssl => true)"
```
robbavey, line 139: Maybe could be reworded a little?

```suggestion
        configuration_error "ssl_certificate => is a required setting when ssl=>true is configured"
```

lib/logstash/inputs/beats.rb
```diff
@@ -131,44 +131,40 @@ def register
...
+      end
 
-    if @ssl && require_certificate_authorities? && !client_authentification?
-      raise LogStash::ConfigurationError, "Using `verify_mode` set to PEER or FORCE_PEER, requires the configuration of `certificate_authorities`"
-    end
+      if require_certificate_authorities? && !client_authentification?
+        configuration_error "Using verify_mode => '#{@ssl_verify_mode}' requires the configuration of ssl_certificate_authorities => ..."
```
robbavey, line 143: Maybe use `ssl_verify_mode` instead of `verify_mode` to be consistent with the configuration name

```suggestion
        configuration_error "ssl_certificate_authorities => is a required setting when ssl_verify_mode => '#{@ssl_verify_mode}' is configured"
```

lib/logstash/inputs/beats.rb
```diff
@@ -131,44 +131,40 @@ def register
...
+      end
 
-    if client_authentication_metadata? && !require_certificate_authorities?
-      raise LogStash::ConfigurationError, "Enabling `peer_metadata` requires using `verify_mode` set to PEER or FORCE_PEER"
+      if client_authentication_metadata? && !require_certificate_authorities?
+        configuration_error "Enabling ssl_peer_metadata => true requires using verify_mode => 'peer' or 'force_peer'"
```
robbavey, line 147: Maybe use `ssl_verify_mode` instead of `verify_mode` to be consistent with the configuration name

```suggestion
        configuration_error "Configuring ssl_peer_metadata => true requires `ssl_verify_mode` to be configured with `peer` or `force_peer`"
```

lib/logstash/inputs/beats.rb
```diff
@@ -177,11 +173,40 @@ def create_server
...
+      server.setSslHandlerProvider(new_ssl_handshake_provider(ssl_context_builder))
     end
     server
   end
 
+  def new_ssl_handshake_provider(ssl_context_builder)
+    begin
+      org.logstash.netty.SslHandlerProvider.new(ssl_context_builder.build_context, @ssl_handshake_timeout)
+    rescue java.lang.IllegalArgumentException => e
+      @logger.error("SSL configuration invalid", :message => e.message)
```
robbavey, line 185: Does it make sense to include the backtrace here? Maybe optionally if `debug` logging is enabled?

lib/logstash/inputs/beats.rb
```diff
@@ -177,11 +173,40 @@ def create_server
...
+  end
+
+  def new_ssl_context_builder
+    passphrase = @ssl_key_passphrase.nil? ? nil : @ssl_key_passphrase.value
+    begin
+      org.logstash.netty.SslContextBuilder.new(@ssl_certificate, @ssl_key, passphrase)
+                                .setProtocols(convert_protocols)
+                                .setCipherSuites(normalized_ciphers)
+    rescue java.lang.IllegalArgumentException => e
+      @logger.error("SSL configuration invalid", :message => e.message)
```
robbavey, line 200: Backtrace useful here?

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -133,4 +133,13 @@ private String format(String message) {
 
         return "[local: " + localhost + ", remote: " + remotehost + "] " + message;
     }
+
+    private static final int MAX_CAUSE_NESTING = 10;
+
+    private static Throwable extractCause(final Throwable ex, final int nesting) {
```
robbavey, line 139: Could also use a slow "pointer"/fast "pointer" to detect exception loops, if an exception loop is common for this kind of error, but this is ok too if not

robbavey (changes requested):
Thanks for doing this! Just a few minor comments to clarify/improve consistency of some of the logging statements

lib/logstash/inputs/beats.rb
kares: thanks, fine by me - what I was trying is to be a bit similar to the generic (required) parameter validation:
```
Missing a required setting for the beats input plugin:

  input {
    beats {
      port => # SETTING MISSING
      ...
    }
  }
```

lib/logstash/inputs/beats.rb
kares: these are plugin 'owned' exceptions but it won't hurt to have a trace with debug logging

src/main/java/org/logstash/beats/BeatsHandler.java
kares: have checked this manually enforcing a TLS certificate validation failure - which gets us 4 levels of nested causes.
at this point I am not sure loops are common just wanted to make sure we do not accidentally end up looping too much here, esp. as these are considered non-fatal and won't halt plugin execution.

imho, what's important is that we try the extraction since the info logging the plugin did previously was not providing useful information on the 'real' cause + also gave 'unclear' logging based on the following condition:

`cause.getMessage() == null ? cause.getClass().toString() : cause.getMessage()`
which made me previously believe we had a `javax.net.ssl.SSLHandshakeException: General SSLEngine problem` around here while in fact it was wrapped in a `io.netty.handler.codec.DecoderException` with `"javax.net.ssl.SSLHandshakeException: General SSLEngine problem"` as the message.

the whole exception stack is later printed by Netty's logger.

robbavey: Makes sense - I only asked about Exception loops because I encountered them working on the JMS plugin, and added a guard against them.

robbavey (approved):
LGTM, assuming travis (eventually) is green

PR #404, opened 2021-01-20
Ecs compatibility

### LINKED ISSUE #387: Implement ECS-Compatibility Mode
_This is a stub issue, and needs to be fleshed out with details specific to
this plugin._

***

As a part of the effort to make plugins able to run in an ECS-Compatible manner
by default in an upcoming release of Logstash, this plugin needs to either
implement an ECS-Compatibility mode or certify that it does not implicitly use
fields that conflict with ECS.

----

This plugin adds two fields into the event:
- `[host]` with hostname brought from `[beat][hostname]` field contained in the Beats event. This is optional and managed by the deprecated `add_hostname`. Also if deprecated it should work on old Logstash, so it should be remapped to `[host][name]`
- `[@metadata][ip_address]` is the remote ip of the client connection, this is moved into `[@metadata][beat][source][ip]`

PR implementing this changes is #404

andsel:
## What does this PR do?
Add ecs compatibility modes (`disabled` and `v1`). In `disabled` nothing changes but in `v1` a couple of fields added by the plugin are renamed:
- `host` becomes ~~`host.name`~~ `[@metadata][input][beats][host][name]` also if this field is managed by deprecated setting `add_hostname`
- `@metadata.ip_address` becomes ~~`@metadata.beat.source.ip`~~ `[@metadata][input][beats][host][ip]`
- `@metadata.tls_peer.protocol` becomes `[@metadata][input][beats][tls][version_protocol]`
- `@metadata.tls_peer.subject` becomes `[@metadata][input][beats][tls][client][subject]`
- `@metadata.tls_peer.cipher_suite` becomes `[@metadata][input][beats][tls][cipher]`

Note that:
`@metadata.tls_peer.status` keeps it's name because there isn't a 1:1 mapping in ECS for the same concept

This enables the user to prepare their pipelines for ECS

Fixes #387


lib/logstash/inputs/beats.rb
```diff
@@ -156,6 +162,10 @@ def register
       configuration_error "Multiline codec with beats input is not supported. Please refer to the beats documentation for how to best manage multiline data. See https://www.elastic.co/guide/en/beats/filebeat/current/multiline-examples.html"
     end
 
+    # define ecs name mapping
+    @field_hostname = ecs_select[disabled: "host",                    v1: "[host][name]"]
+    @field_hostip   = ecs_select[disabled: "[@metadata][ip_address]", v1: "[@metadata][beat][source][ip]"]
```
kares, line 167: shouldn't this be rather at least a plural `[@metadata][beats][source][ip]` or `[input]` prefixed?

maybe it would be good to have a clear name-space convention used everywhere.
e.g. if there's input-http and filter-http then using `[@metadata][http][url][original]` might collide, thoughts?

spec/inputs/beats/message_listener_spec.rb
```diff
@@ -52,10 +52,74 @@ def flush(&block)
...
+
+  let(:message) { MockMessage.new("abc",  data)}
+
+  it "extract the event" do
+    subject.onNewMessage(ctx, message)
+    event = queue.pop
+    expect(event.get("message")).to be_nil
+    expect(event.get("metric")).to eq(1)
+    expect(event.get("name")).to eq("super-stats")
+    puts "DNADBG>> ecs_compatibility: #{ecs_compatibility} host_field_name: #{host_field_name}, event: #{event.to_hash}"
```
kares, line 79: :scissors:

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -31,7 +31,12 @@ def onNewMessage(ctx, message)
       hash = message.getData
       ip_address = ip_address(ctx)
 
-      hash['@metadata']['ip_address'] = ip_address unless ip_address.nil? || hash['@metadata'].nil?
+      unless ip_address.nil? || hash['@metadata'].nil?
+        tmp_evt = org.logstash.Event.new
+        tmp_evt.setField(@input.field_hostip, ip_address)
+        # Leverage Logstash's event to convert nested event fields into nested hashes
+        nested_fields_to_maps(hash, tmp_evt)
```
kares, line 38: :thinking: - turned out to be a bit complex, but I guess it does what it needs to do

kares (commented):
slight concern on having consistent @metadata "name-spacing" across plugins but otherwise :+1:

lib/logstash/inputs/beats/message_listener.rb
andsel: I think the same, but the alternative was to replicate to logic to unpack nested fields names into maps with essentially replicating the behavior inside the plugin, that was even worst since we introduced duplication.

lib/logstash/inputs/beats.rb
andsel: Reading at https://www.elastic.co/guide/en/ecs/current/ecs-host.html probably one place for `ip` and `name` are inside `host`, following the description "should be populated with details about the host on which the event happened," effectively the host where the event is happening is the host that's running the Beats process.

kares: that's a different discussion but a good point - LS being a bridge here for the Beats agent it does not make much sense to set `host.name`. maybe set it as a meta-data fiels for both cases in ECS mode as : 
- `[@metadata][input][beats][host][name]`
- and `[@metadata][input][beats][host][ip]`

... users could still use the field in the pipeline config and even copy the field if they need it but our advice would be to not fill `host` in ECS mode?

andsel: From that perspective it sounds reasonable, however the `@metadata` are not forwarded to the downstream systems (like ES) but if the user want it could just rename it to something it consider usable in his use case, respecting the ECS.
Given this, also the idea to separate these metadata by `plugin_type` and `plugin_name`, like you suggested in the case of `[input][http]` and `[filter][http]` seems correct

docs/index.asciidoc
```diff
@@ -144,6 +152,29 @@ The list of ciphers suite to use, listed by priorities.
...
+Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)].
+The value of this setting affects the placement of a TCP connection's metadata on events:
+
+.Metadata Location by `ecs_compatibility` value
+[cols="<l,<l,e,<e"]
+|=======================================================================
+|`disabled`              |`v1`                      |Availability |Description
+
+|[host]                  |[@metadata][input][beats][host][name]              |Always       |Name or address of the beat host
+|[@metadata][ip_address] |[@metadata][source][ip]   |Always       |IP address of source end of the connection
```
kares, line 175:
```suggestion
|[@metadata][ip_address] |[@metadata][input][beats][source][ip]   |Always       |IP address of of the Beats client
```

lib/logstash/inputs/beats/decoded_event_transform.rb
```diff
@@ -9,7 +9,16 @@ def transform(event, hash)
       ts = coerce_ts(hash.delete("@timestamp"))
 
       event.set("@timestamp", ts) unless ts.nil?
-      hash.each { |k, v| event.set(k, v) }
+      hash.each do |k, v|
+        #could be a nested map, so we need to merge and not overwrite
+        existing_value = event.get(k)
+        if (!existing_value.nil? && existing_value.is_a?(Hash))
```
kares, line 15:
```suggestion
        if existing_value.is_a?(Hash)
```

docs/index.asciidoc
```diff
@@ -144,6 +152,29 @@ The list of ciphers suite to use, listed by priorities.
...
+* Value type is <<string,string>>
+* Supported values are:
+** `disabled`: unstructured connection metadata added at root level
+** `v1`: structured connection metadata added under ECS compliant namespaces
+* Default value depends on which version of Logstash is running:
+** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
+** Otherwise, the default value is `disabled`.
+
+Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)].
+The value of this setting affects the placement of a TCP connection's metadata on events:
```
kares, line 167:
```suggestion
The value of this setting affects the keys for Beats connection's meta-data on the event:
```
// cc @karenzone - please take a look if the sentence makes sense

docs/index.asciidoc
```diff
@@ -86,6 +86,13 @@ Logstash `@timestamp` field.
 This configuration results in daily index names like
 +filebeat-{logstash_version}-{localdate}+.
 
+==== Event Metadata and the Elastic Common Schema (ECS)
+This plugin other than decoding `beat` events, adds some fields related to the deprecated `host`,
+containing the `hostname` provided by beats and `ip_address` which contains the remote address
+of the client's connection. These two fields are now moved in ECS compatible namespaces
+when the select compatibility mode is enabled.
+For a complete listing of metadata shape, see <<plugins-{type}s-{plugin}-ecs_compatibility>>.
```
kares, line 94:
```suggestion
This plugin, other than decoding events, adds fields related to the Beats client connection, namely the source host name and ip address. These fields use different names depending on whether ECS compatibility mode is enabled, see <<plugins-{type}s-{plugin}-ecs_compatibility>>.
```
// cc @karenzone please :eyes:

kares (commented):
:+1: looking good, 

docs seemed a bit unclear so I tried to rephrase but would rather wait for a second set of :eyes:

spec/inputs/beats/message_listener_spec.rb
kares: I do feel your pain, think we might have "over-engineered" the specs with the ecs_compatibility meta stuff.
let's see how it plays out in the long run, personally rather have duplicity in specs than having to :thought_balloon: on how it works...

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -31,7 +31,12 @@ def onNewMessage(ctx, message)
...
+      end
       target_field = extract_target_field(hash)
```
kares, line 40: this line has an effect of setting more `@metadata`, 
since we now use a common prefix for the other meta-data `[input][beats][...]`
we should consider also changing https://github.com/logstash-plugins/logstash-input-beats/blob/master/lib/logstash/inputs/beats/message_listener.rb#L143-L152 to use a prefix in ECS mode.

kares (commented):
:dog: looking good, there's a concern about changing `[tls_peer][status] => "verified"`.

if we're changing the other TLS fields I think it should be changed as well, this might be a pain for users (e.g. if they use it in pipeline definition `if [@metadata][tls_peer][status] = "verified"` to enhance events) but with proper documentation on the changes such as the provided name mapping table I think it's acceptable.

spec/support/shared_examples.rb
```diff
@@ -21,12 +21,18 @@
     }
   end
 
+  def key_as_nested_maps(key, value)
+    evt = LogStash::Event.new
+    evt.set(key, value)
+    evt.to_hash_with_metadata
```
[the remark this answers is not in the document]
andsel: If the `key` is in the form `[@metadata][bar]` we want that the hash return contains `{"@metadata" => {"bar" => <value>}}` so I don't understand why you are suggesting to remove all the root level keys that start with `@`?

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -31,7 +31,9 @@ def onNewMessage(ctx, message)
       hash = message.getData
       ip_address = ip_address(ctx)
 
-      hash['@metadata']['ip_address'] = ip_address unless ip_address.nil? || hash['@metadata'].nil?
+      unless ip_address.nil? || hash['@metadata'].nil?
+        set_nested(hash, @input.field_hostip, ip_address)
+      end
       target_field = extract_target_field(hash)
 
       extract_tls_peer(hash, ctx)
```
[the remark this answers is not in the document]
andsel: I understand your intention to keep explicit what's done for the various `ecs_compatibility` levels, but this replicate the behavior and we miss the meaning of `ecs_select`. I thought that the spirit of `ecs_select` is to keep in one place the mapping correspondence between ecs `:disabled` and `:v1`, the same way of using it is also present in `logstash-input-udp` and I think that keep the same style across plugins would be helpful.

I've some doubts about `['@metadata']['tls_peer']` proposed suggestions:
- could we change the type of a field when we rename it? For example "verified/unverified" for `:disabled` ECS becomes `true/false` when `:v1` is enabled. This change would require more changes in user's pipelines during the transition. 
- the `tls` section is described in [docs](https://www.elastic.co/guide/en/ecs/current/ecs-tls.html) and doesn't specify any `verified` field, so could we add a custom named field nesting into an already ECS defined structure? Do we risk conflicts with future versions? Could be this confusing?

spec/support/shared_examples.rb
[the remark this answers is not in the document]
kares: my bad, was only reading non `@metadata` usage - and wanted to avoid merging `@timestamp` to the original event

lib/logstash/inputs/beats/message_listener.rb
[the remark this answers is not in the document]
kares: :+1: believe there's a misunderstanding here.
my original suggestion to review `hash['@metadata']['tls_peer']` in the context of unifying the `[input][beats]` prefix was simply to avoid a non-namespaced hash - which we can not achieve if `['@metadata']['tls_peer']['status']` has to stay.

if we need to keep it than I am less concerned about the need to change the rest but I now see that it can still provide some value if users extract the whole `[@metadata][input][beats][tls]` hash they get ECS compliant TLS information. 

> we miss the meaning of `ecs_select`

true in terms of consistency - still it hurt my :eyes: in terms of readability as there are many fields we set at once ...
not sure it's improving the code here but I am willing to :peace_symbol: with it.

> the tls section is described in docs and doesn't specify any `verified` field, so could we add a custom named field nesting into an already ECS defined structure?

was fully aware `[tls][verified]` not being compliant, but there's no clash since there's a custom namespace in front (+ it's meta-data). again my intention was the removal of `tls_peer` hash as we started introducing a convention for plugins to use a prefix convention. the risk of 'breaking' user pipelines initially seemed worth having a clean slate in ECS mode.

not breaking pipelines is more important than being "logically consistent" - **feel free to disregard this review part**.

kares (approved):
:+1: no further concerns on my end

the manual `FieldReference.from` usage looks ugly but there's no other way atm to handle nested event hash merges, hopefully we can have something in LS itself to avoid these ...

docs/index.asciidoc
```diff
@@ -144,6 +151,33 @@ The list of ciphers suite to use, listed by priorities.
...
+
+* Value type is <<string,string>>
+* Supported values are:
+** `disabled`: unstructured connection metadata added at root level
+** `v1`: structured connection metadata added under ECS compliant namespaces
+* Default value depends on which version of Logstash is running:
+** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
+** Otherwise, the default value is `disabled`.
+
+Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)].
```
karenzone, line 165:
```suggestion
Controls this plugin's compatibility with the {ecs-ref}[Elastic Common Schema (ECS)].
```

docs/index.asciidoc
```diff
@@ -86,6 +86,12 @@ Logstash `@timestamp` field.
 This configuration results in daily index names like
 +filebeat-{logstash_version}-{localdate}+.
 
+==== Event Metadata and the Elastic Common Schema (ECS)
```
karenzone, line 89:
```suggestion

[id="plugins-{type}s-{plugin}-ecs_metadata"]
==== Event Metadata and the Elastic Common Schema (ECS)
```

docs/index.asciidoc
```diff
@@ -144,6 +151,33 @@ The list of ciphers suite to use, listed by priorities.
...
+* Value type is <<string,string>>
+* Supported values are:
+** `disabled`: unstructured connection metadata added at root level
+** `v1`: structured connection metadata added under ECS compliant namespaces
+* Default value depends on which version of Logstash is running:
+** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
+** Otherwise, the default value is `disabled`.
+
+Controls this plugin's compatibility with the https://www.elastic.co/guide/en/ecs/current/index.html[Elastic Common Schema (ECS)].
+The value of this setting affects the keys for Beats connection's meta-data on the event:
```
karenzone, line 166:
```suggestion
The value of this setting affects the keys for the Beats connection's metadata on the event:
```

docs/index.asciidoc
```diff
@@ -86,6 +86,12 @@ Logstash `@timestamp` field.
 This configuration results in daily index names like
 +filebeat-{logstash_version}-{localdate}+.
 
+==== Event Metadata and the Elastic Common Schema (ECS)
+This plugin other than decoding `beat` events, adds some fields related to the deprecated `host`,
+containing the `hostname` provided by beats and `ip_address` which contains the remote address
+of the client's connection. These two fields are now moved in ECS compatible namespaces
+when the select compatibility mode is enabled, see <<plugins-{type}s-{plugin}-ecs_compatibility>>.
+
```
karenzone, line 94: What about something like this: 

> When <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is
enabled, this plugin adds {ecs-ref}[ECS] fields containing the `hostname` provided by beats and
the remote `ip_address` of the client's connection.

karenzone (commented):
I added some docs suggestions inline. Pls let me know if you'd like to discuss.

docs/index.asciidoc
andsel: I think that the meaning is shifted, from the sentence above seems like that the 2 fields are added only when ecs is enabled, but the plugin always add those fields, the only difference is that when ecs is enabled those fields are remapped

andsel: I tried to rephrase it as:

> This plugin other than decoding `beat` events adds a couple of fields related to the event, the deprecated `host`
> which contains the `hostname` provided by beats and the `ip_address` containing the remote address
> of the client's connection. When <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is
> enabled these are now moved in ECS compatible namespaces.


LMKWDYT

karenzone: Thanks for the explanation. 
Would you consider changing the introduction slightly, to "In addition to decoding..."  or "When decoding..."

>When decoding `beats` events, this plugin adds two fields related to
the event: the deprecated `host` which contains the `hostname` provided by beats
and the `ip_address` containing the remote address of the client's connection.
When <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is
enabled these are now moved in ECS compatible namespaces.

karenzone: Thanks for adding the other suggestions. Everything else looks as expected.

karenzone (approved):
Re-tested. Builds cleanly and LGTM
Thanks for your work on this

PR #405, opened 2021-01-26
Feat: log + unwrap generic SSL context exceptions 

kares:
The unwrapping at the Java level is for [exceptions wrapped by Netty](https://github.com/netty/netty/blob/netty-4.1.49.Final/handler/src/main/java/io/netty/handler/ssl/JdkSslServerContext.java#L288).

These exceptions as they are are rather useless (unless they're properly logged with a cause's `printStackTrace`), and since we're pretty much type-less Ruby on the layer above we should rather propagate the cause ...

Full exception details will be logged at debug level from the Java side - since we seem to prefer (manual) exception logging at the plugin level. We also make sure to log cause, if any, on the Ruby side which now catches all (expected) Java exceptions.


andsel (approved):
LGTM

PR #409, opened 2021-02-12
Add documentation for executor_threads

jsvd:


jsvd:
this may be unnecessary with #410

karenzone:
Looks like this work was subsumed in #421. Should we close this one?

PR #412, opened 2021-03-29
User super() instead of super in RSpec shared examples

andsel:
Commit 3f03ea37b7e101cfd36de880bcba2eb6d40f28cf switched the usage of `super` to `super()` this made fail some tests on 8.0.0-SNAPSHOT CI:  https://github.com/logstash-plugins/logstash-input-beats/runs/2205409837.

This PS fixes the failing tests


kares (approved):
:green_circle:

PR #413, opened 2021-03-29
Doc: Move ECS table to the Initial section

roaksoax:
Moves the ECS mapping table to the initial section that describes event metadata and the event.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -92,7 +92,7 @@ This configuration results in daily index names like
 When decoding `beats` events, this plugin adds two fields related to the event: the deprecated `host`
 which contains the `hostname` provided by beats and the `ip_address` containing the remote address
 of the client's connection. When <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is
-enabled these are now moved in ECS compatible namespace:
+enabled these are now moved in ECS compatible namespace. Here's how <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> affects output.  
 
 [cols="<l,<l,e,<e"]
 |=======================================================================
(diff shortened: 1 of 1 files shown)


docs/index.asciidoc
```diff
@@ -92,7 +92,19 @@ This configuration results in daily index names like
 When decoding `beats` events, this plugin adds two fields related to the event: the deprecated `host`
 which contains the `hostname` provided by beats and the `ip_address` containing the remote address
 of the client's connection. When <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is
-enabled these are now moved in ECS compatible namespace.
+enabled these are now moved in ECS compatible namespace:
```
karenzone, line 95:
```suggestion
enabled these are now moved in ECS compatible namespace. Here's how <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> affects output.  
```

karenzone (approved):
Nice! I like the mapping info grouped with the higher level conceptual info. I suggested a minor change to the sentence before the table for your consideration.  

Otherwise, builds cleanly and LGTM.

docs/index.asciidoc
```diff
@@ -92,7 +92,19 @@ This configuration results in daily index names like
...
+
+[cols="<l,<l,e,<e"]
+|=======================================================================
+|`disabled`              |`v1`                      |Availability |Description
```
andsel, line 99: As a minor change, I would remove the backticks because are rendered as character and do not change the appearance of `disable` and `v1` strings.

andsel (approved):
LGTM

karenzone:
Here's how it looks:
Screen Shot 2021-03-30 at 12 59 34 PM

I liked the back ticks, even though they weren't rendering. They provided a visual indication that "disabled" and "v1" were config options. If everybody else prefers it this way, I'll live with it.  :-)  LGTM

PR #415, opened 2021-03-30
Doc: Add plugin name attribute

karenzone:
Related: https://github.com/elastic/logstash/issues/12794

Adds and expands use of name attributes to facilitate Beats/Agent doc support.
Replace hard-coded instances of plugin name with attribute (where it makes sense).  This work will continue over time.


andsel (approved):
LGTM

karenzone:
Published at v.6.1.2: https://rubygems.org/gems/logstash-input-beats/versions/6.1.2-java

PR #416, opened 2021-04-14
Doc: Remove unnecessary file

karenzone:
This source file is no longer needed after we added variables to the shared source file in #415 

No version bump is necessary.


andsel (approved):
LGTM

PR #417, opened 2021-04-15
Doc: Add variable to code sample

karenzone:
In #415 we implemented attributes and set up code samples to accommodate attributes, but didn't actually implement the attribute in the code sample. This PR switches out the hard coded value in the code sample with an attribute. 

**ToDo:**
*  Look through source file to be sure we didn't miss any other instances of hard coded plugin name. 
*  Line 49 "<1> `%{[@metadata][beat]}` sets the first part of the index name to the value
of the `beat` metadata field" . Figure out best implementation for `beat` (singular).
* Determine messaging and location for relationship of origin/destination of aliased plugins. (See an idea in comments below.) 

**Note:**
* An across the board search-replace won't work. We have to be selective to avoid screwing up context and breaking links.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -18,6 +18,9 @@ END - GENERATED VARIABLES, DO NOT EDIT!
 
 === {plugin-uc} input plugin
 
+NOTE: The `input-elastic_agent` plugin is the next generation of the
+`input-beats` plugin. They currently share a common codebase.
+
 include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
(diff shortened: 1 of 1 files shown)


karenzone:
An idea for communicating alias info: 
Screen Shot 2021-04-21 at 7 03 34 PM

Location and messaging still TBD. Comments welcomed.

karenzone:
@andsel  I'm making more changes per #422 (as we discussed) in a subsequent PR.

andsel (approved):
LGTM

Just a note on 
> Line 49 "<1> %{[@metadata][beat]} sets the first part of the index name to the value
of the beat metadata field"

the alias remapping works only on the rename of the config section but being the code the same, it doesn't rename `@metadata`'s sub field. In that case we need to adapt the plugin's code

Issue #408, opened 2021-02-04
safe quard potential buffer allocation


kares:
managed to get a Netty warning on a [buffer.release] after letting Beats send logs for a few hours :
```
[2021-02-02T17:54:49,050][ERROR][io.netty.util.ResourceLeakDetector][main][beats-tls] LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newHeapBuffer(PooledByteBufAllocator.java:346)
	io.netty.buffer.AbstractByteBufAllocator.heapBuffer(AbstractByteBufAllocator.java:168)
	io.netty.buffer.AbstractByteBufAllocator.heapBuffer(AbstractByteBufAllocator.java:159)
	org.logstash.beats.BeatsParser.inflateCompressedFrame(BeatsParser.java:218)
	org.logstash.beats.BeatsParser.decode(BeatsParser.java:182)
	io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:501)
	io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:440)
	io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:61)
	io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:370)
	io.netty.util.concurrent.DefaultEventExecutor.run(DefaultEventExecutor.java:66)
	io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	java.lang.Thread.run(Thread.java:748)
```
[...]

PR #420, opened 2021-04-20
Fix: safe-guard byte buf allocation

kares:
since the `inflateCompressedFrame` might potentially leak, see https://github.com/logstash-plugins/logstash-input-beats/issues/408


yaauie (approved):
LGTM 👍🏼 

This is an elegant solution, ensuring that the affected `buffer`'s whole lifecycle is managed inside `inflateCompressedFrame`.

PR #428, opened 2021-07-22
Doc: Remove link to limitations topic

karenzone:
As Elastic Agent matures, limitations are being removed.  The topic targeted from the link will be going away, and so the link must be deleted, too.


karenzone:
We need the version bump/publish to make sure that we're picking up correct content when we generate plugins for 7.x and master

andsel (approved):
LGTM

PR #429, opened 2021-07-28
ecs: add v8 alias to v1 implementation

yaauie:
Adds alias `v8` pointing to the existing `v1` implementation, since there are no known relevant divergences in the upcoming ECS 8.0 release.


kares (approved):
:strawberry:  thanks for adding event_factory support

PR #442, opened 2021-12-30
Test: refactor setup and get all tests passing

kares:
back to :green_circle: 
- adjustment due LS switching the base image CentOS 7 -> Ubuntu 20.04
  harder than expected due #443 
- fix a :red_circle: due test assuming ECS disabled default

refactored the *Dockerfile* to use the base file (from **.ci**) and have the extra `install openssl` in `.ci/setup.sh` hook


spec/inputs/beats_spec.rb
```diff
@@ -166,7 +166,14 @@
   end
 
   context "tls meta-data" do
-    let(:config) { super().merge("host" => host, "ssl_peer_metadata" => true, "ssl_certificate_authorities" => [ certificate.ssl_cert ]) }
+    let(:config) do
+      super().merge(
+          "host" => host,
+          "ssl_peer_metadata" => true,
+          "ssl_certificate_authorities" => [ certificate.ssl_cert ],
+          "ecs_compatibility" => 'disabled'
```
kaisecheng, line 174: we don't have a test case for tls metadata in ecs mode, isn't it?

kaisecheng (approved):
The change LGTM. Thanks for fixing the CI.
A suggestion is to add a test case for tls metadata in ecs mode.

spec/inputs/beats_spec.rb
kares: might be elsewhere - recall writing this spec as specifically guarding against a [NPE regression](https://github.com/logstash-plugins/logstash-input-beats/pull/431), 
the bug surfaced regardless of `ecs_compatibility` and the test here isn't about verifying ECS vs non-ECS field behavior.

yaauie (approved):

PR #447, opened 2022-03-10
Feat: add support for TLSv1.3

tsaarni:
This change adds support for TLSv1.3.

Updates https://github.com/elastic/logstash/issues/10494

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

Open questions:

- [x] JCE limited strength jurisdiction ciphers are not updated by this PR.  Find out if TLSv1.3 ciphers can be added there.
- [x] Is Java 8 compatibility needed? If not, then `TLS_CHACHA20_POLY1305_SHA256` can be added to TLSv1.3 ciphers.
- [x] Decide if the cipher auto-detection for CHA-CHA stays, if yes, describe it in the docs.


tsaarni:
@kares wrote (originally from https://github.com/elastic/logstash/issues/10494#issuecomment-1066621216)
> @tsaarni thanks for the effort, whatever progress you're able to do I can built upon, what might server as an example to some extent is what we did in TCP input ... as to how much existing Beats option we'll deprecate in favor of a naming similar to the TCP input added options is undecided at this point.

For now I have updated currently existing options `tls_max_version` and  `cipher_suites` and did not change/deprecate the values. Would this be acceptable approach at this point?

One test job failed but maybe it is flake since the error is `Error response from daemon: manifest for docker.elastic.co/logstash/logstash:8.1.1-SNAPSHOT not found: manifest unknown: manifest unknown`

src/main/java/org/logstash/netty/SslContextBuilder.java
```diff
@@ -52,7 +52,9 @@
             "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384",
             "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384",
             "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",
-            "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256"
+            "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",
+            "TLS_AES_128_GCM_SHA256",
+            "TLS_AES_256_GCM_SHA384",
```
kares, line 57: believe these should be at the top (still unsure about trying to detect the cha-cha ciphers)

src/main/java/org/logstash/netty/SslContextBuilder.java
```diff
@@ -68,7 +70,7 @@
     private String[] supportedCiphers = ((SSLServerSocketFactory)SSLServerSocketFactory
             .getDefault()).getSupportedCipherSuites();
     private String[] ciphers = DEFAULT_CIPHERS;
-    private String[] protocols = new String[] { "TLSv1.2" };
+    private String[] protocols = new String[] { "TLSv1.3" };
```
kares, line 73: for backwards compatibility v1.3 should not be forced, having `new String[] { "TLSv1.2", "TLSv1.3" };` should provide backwards compatibility with enabling TLS 1.3 as well

kares (commented):
:+1: on track - left a few comments on code changes

I am planning to revisit this with some kind of a TLSv1.3 "happy path" test

kares:
> JCE limited strength jurisdiction ciphers are not updated by this PR. Find out if TLSv1.3 ciphers can be added there.

hopefully we do not need to care about that part anymore - will need to double check. fine to leave it as is for now.

> Is Java 8 compatibility needed? If not, then `TLS_CHACHA20_POLY1305_SHA256` can be added to TLSv1.3 ciphers.

unfortunately Java 8 compat is still needed - wonder if we should do some kind of 'light' feature detection of the cipher being available ...

src/main/java/org/logstash/netty/SslContextBuilder.java
tsaarni: Fixed.  

I also added proposal for auto-detecting cha-cha: All the ciphers are now in `DEFAULT_CIPHERS` but when `getDefaultCiphers()` is called it returns union of default and what is supported by current Java version.  I guess documentation should then have two versions of defaults (not updated yet)?

I added some kind of minimal test case for that too and noticed that the existing TLS Java tests assumed certificate in `/tmp`, so I added one test cert in test resources from [TCP input fixtures](https://github.com/logstash-plugins/logstash-input-tcp/tree/main/spec/fixtures).

I can remove this auto-detect stuff if it does not sound like a good idea after all, I don't mind!

src/main/java/org/logstash/netty/SslContextBuilder.java
tsaarni: Fixed.

tsaarni:
Not sure why `bundle install` seems failing https://app.travis-ci.com/github/logstash-plugins/logstash-input-beats/jobs/563503327#L518

src/main/java/org/logstash/netty/SslContextBuilder.java
kares: >  also added proposal for auto-detecting cha-cha: All the ciphers are now in `DEFAULT_CIPHERS` but when `getDefaultCiphers()` is called it returns union of default and what is supported by current Java version.

believe, this is fine for now - some kind of filtering as the best we could do anyways.

> I guess documentation should then have two versions of defaults (not updated yet)?

I think we can have just one set not to confuse users - let's include the CHA-CHA ciphers and have a note in the documentation that mentions that the list is accurate for Java 11+

kares:
> I went ahead and added a happy path test in integration suite, but let me know if this is something you were thinking. I noticed that suite had old filebeat that did not support TLSv1.3 so I needed to upgraded it. Went for latest but 7.x would be OK as well.

:100: - great stuff Tero, regarding 7.x it would be better to have the older version as the plugin tries to maintain backwards compatibility 7.8 was supported a few months back but any Beats from 7.8 - 7.10 range that has TLS 1.3 and works would be for the best. I do not mind having 8.x as well but in that case we would need to take things further and parameterize the version of Beats we're testing against (and have different CI targets).

> Also small change in `spec/support/file_helpers.rb`: Filebeat was not happy with too wide config file permissions when executed in my machine. The files created via `write_to_tmp_file()` have now forced 0600 permissions, not depending on system's umask to be set accordingly.

fine by me to have narrower configuration - hopefully it still works with 7.x

tsaarni:
> regarding 7.x it would be better to have the older version as the plugin tries to maintain backwards compatibility 7.8 was supported a few months back but any Beats from 7.8 - 7.10 range that has TLS 1.3 and works would be for the best. 

Filebeat 7.6.0 was the first one with TLS 1.3.  It seemed to work, except for the errors due to bundler problem https://github.com/elastic/logstash/issues/13890.

kares:
We just pushed a plugin update with a similar code base: https://github.com/logstash-plugins/logstash-input-http/pull/146

There's a couple of things that are handled slightly differently [(loading the cipher lists ahead of time)](https://github.com/logstash-plugins/logstash-input-http/pull/146/files#diff-cd477b6c3d223bfe35f4076bec10e6240dbfaf88d5d714c91e90e80e93a93700R29-R57), could you take a look and bump the version number and include a CHANGELOG entry please?

tsaarni:
> There's a couple of things that are handled slightly differently [(loading the cipher lists ahead of time)](https://github.com/logstash-plugins/logstash-input-http/pull/146/files#diff-cd477b6c3d223bfe35f4076bec10e6240dbfaf88d5d714c91e90e80e93a93700R29-R57), could you take a look and bump the version number and include a CHANGELOG entry please?

Sure! I have now copied the default/supported ciphers handling from https://github.com/logstash-plugins/logstash-input-http/pull/146.

I stepped minor version number like it seemed to be done in HTTP input plugin as well.

kares (commented):
:+1: looking great, thanks again for the contribution!

Planning to run some manual testing with Beats next week and will release afterwards unless I bump into any hiccups.

kares (approved):
Tested with Beats (7.16) and all seems to be working:

- when TLS 1.3 was forced on Beats' end things already work even wout this PR (as expected)
- when `tls_min_version => 1.3` was able to get Beats to work when `ssl.supported_protocols: [TLSv1.3]` (or simply when `ssl.supported_protocols` is not set)
- with `tls_min_version => 1.3` and `ssl.supported_protocols: [TLSv1.2]` Beats failed as expected

PR #449, opened 2022-04-18
Remove use of deprecated java `import` from ruby

robbavey:
As of jruby-9.3, the use of `import` in ruby file to import java classes
has been deprecated, and the use of `java_import` is recommended.

Relates: https://github.com/jruby/jruby/issues/6976, https://github.com/jruby/jruby/pull/7143


jsvd (approved):
LGTM

PR #450, opened 2022-04-19
Feat: review and deprecate ssl protocol/cipher settings

kares:
> ... same as https://github.com/logstash-plugins/logstash-input-http/pull/151

(a follow-up on https://github.com/logstash-plugins/logstash-input-beats/pull/447)

- deprecating `cipher_suites` in favor of `ssl_cipher_suites` (with better validation of the supported cipher set)
- adding `ssl_supported_protocols` as a replacement for `tls_min_version` / `tls_max_version`


andsel (approved):
LGTM!

docs/index.asciidoc
```diff
@@ -160,7 +160,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 |=======================================================================
 |Setting |Input type|Required
 | <<plugins-{type}s-{plugin}-add_hostname>> |<<boolean,boolean>>|No
```
karenzone, line 162:
```suggestion
| <<plugins-{type}s-{plugin}-add_hostname>> |<<boolean,boolean>>|__Deprecated__
```

docs/index.asciidoc
```diff
@@ -217,14 +214,14 @@ Close Idle clients after X seconds of inactivity.
...
-** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
-** Otherwise, the default value is `disabled`.
+  * Value type is <<string,string>>
+  * Supported values are:
+    ** `disabled`: unstructured connection metadata added at root level
+    ** `v1`: structured connection metadata added under ECS v1 compliant namespaces
+    ** `v8`: structured connection metadata added under ECS v8 compliant namespaces
+  * Default value depends on which version of Logstash is running:
+    ** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
+    ** Otherwise, the default value is `disabled`.
```
karenzone, line 224: The asciidoc formatting is correct as it was.  You don't even have to add the leading space to the first level bullets. Asciidoc automatically indents them for you. 
The second level bullets are tagged `**` and should be aligned with the left margin.

docs/index.asciidoc
kares: meh, nice catch. thanks!

docs/index.asciidoc
```diff
@@ -175,8 +175,9 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-ssl_key_passphrase>> |<<password,password>>|No
 | <<plugins-{type}s-{plugin}-ssl_verify_mode>> |<<string,string>>, one of `["none", "peer", "force_peer"]`|No
```
karenzone, line 176: While we're making things right, will you please move the `ssl_verify_mode` option to its proper place in alpha order--both here and in the description?

karenzone (commented):
Left some comments inline. Otherwise, LGTM!

docs/index.asciidoc
kares: yy it just read weird - the rest of the documented options had the whitespaces to aid while reading the source

karenzone: If you like the leading spaces before the first level bullets, I don't have a problem with that. They won't hurt anything. 

Asciidoc standard is to left align all bullets. I like following the standards because otherwise, people see things done differently and a new "pseudo standard" is born.

karenzone (commented):
Left comment for consideration. Otherwise, LGTM.

PR #454, opened 2022-08-11
[Doc] add memory suggestion

kaisecheng:
Many users come across `java.lang.OutOfMemoryError: Direct buffer memory`
This commit adds a memory session to set the expectation of how much memory the plugin could use.


docs/index.asciidoc
```diff
@@ -88,6 +88,14 @@ will be similar to events directly indexed by {plugin-uc} into Elasticsearch.
 endif::[]
 
 
+[id="plugins-{type}s-{plugin}-memory"]
+===== Memory usage
+
+This plugin uses direct memory which is not included in the heap memory.
+By default, it is the same as the heap size (Xmx), for example, setting 10GB Xmx means setting 10GB of direct memory.
```
yaauie, line 95:
```suggestion
By default, a JVM's off-heap direct memory limit is the same as the heap size.
For example, setting a heap size of 10GB without setting the direct memory limit will result in the direct limit _also_ being 10GB for a total of 20GB.
```

docs/index.asciidoc
```diff
@@ -88,6 +88,14 @@ will be similar to events directly indexed by {plugin-uc} into Elasticsearch.
...
+Make sure heap and direct memory does not exceed the total size of the server. You can set the amount of direct memory with -XX:MaxDirectMemorySize.
```
yaauie, line 96:
```suggestion
Make sure heap and direct memory combined does not exceed the total memory available on the server.
You can set the amount of direct memory with `-XX:MaxDirectMemorySize` in {logstash-ref}/jvm-settings.html[Logstash's JVM Settings].
```

We should also add an off-heap section to [Logstash's JVM Settings](https://github.com/elastic/logstash/blob/main/docs/static/config-details.asciidoc).

karenzone, line 96:
```suggestion
This plugin uses direct memory (bypassing the CPU) to improve performance. 
Direct memory and heap size (Xmx) allocations are separate, but are the same by default. 
For example, setting 10GB Xmx sets direct memory to 10GB.
You can change the amount of direct memory with -XX:MaxDirectMemorySize.

NOTE: Be sure that heap and direct memory total allocation does not exceed the total size of the server.
```

karenzone (commented):
Left some suggestions inline for consideration.

docs/index.asciidoc
```diff
@@ -88,6 +88,14 @@ will be similar to events directly indexed by {plugin-uc} into Elasticsearch.
 endif::[]
 
 
+[id="plugins-{type}s-{plugin}-memory"]
+===== Memory usage
+
+This plugin uses direct memory which is not included in the heap memory.
```
yaauie, line 94:
```suggestion
This plugin uses "off-heap" direct memory _in addition to_ heap memory.
```

yaauie (commented):
I would like for this doc (and/or for the suggested doc addition in LS-core) to cover the _consequences_ of having direct memory limit too low or too high. From my understanding, setting it too high will cause a runtime failure to allocate the direct memory, but if we set it too low do we merely not buffer as much from the inbound connections?

docs/index.asciidoc
robbavey, line 96: Maybe add a link to the [jvm settings](https://www.elastic.co/guide/en/logstash/current/jvm-settings.html) doc in logstash core?

We should also update that document too, along the lines of this [issue](https://github.com/elastic/logstash/issues/12622)

docs/index.asciidoc
robbavey, line 95:
```suggestion
This plugin uses direct memory which is allocated in addition to that allocated to the heap.

By default, this is set to the same value as allocated to the heap via the `Xmx` flags. For example, setting `-Xmx10G` will allocate `10GB` for heap and an additional `10GB` for direct memory, for a total of `20GB` allocated.  
```

robbavey (commented):
Couple of suggestions, but final approval should go to @karenzone or @kilfoyle

kaisecheng:
> I would like for this doc (and/or for the suggested doc addition in LS-core) to cover the _consequences_ of having direct memory limit too low or too high. From my understanding, setting it too high will cause a runtime failure to allocate the direct memory, but if we set it too low do we merely not buffer as much from the inbound connections?

@yaauie Logstash can start with `-XX:MaxDirectMemorySize=100G` in my mac, probably will throw OutOfMemoryError later when it takes more than the total size of the machine. Also, tried to ingest 10G data from filebeat to logstash with `-XX:MaxDirectMemorySize=64M` without exception. I will add a note that set too low direct memory will lower the performance.

docs/index.asciidoc
kaisecheng: Adding a link sounds good. Users mostly come across direct memory with beats-input. I will update the JVM settings page. Thanks for the pointer

kaisecheng:
@karenzone Thanks for the suggestion. I have combined everyone's ideas into one. Also, I update the JVM session in logstash core https://github.com/elastic/logstash/pull/14431 Please review both PRs.

docs/index.asciidoc
```diff
@@ -88,6 +88,17 @@ will be similar to events directly indexed by {plugin-uc} into Elasticsearch.
 endif::[]
 
 
+[id="plugins-{type}s-{plugin}-memory"]
+===== Memory usage
+
+This plugin uses "off-heap" direct memory in addition to heap memory.
+By default, a JVM's off-heap direct memory limit is the same as the heap size.
+For example, setting `-Xmx10G` without setting the direct memory limit will allocate `10GB` for heap and an additional `10GB` for direct memory, for a total of `20GB` allocated. 
+You can set the amount of direct memory with `-XX:MaxDirectMemorySize` in {logstash-ref}/jvm-settings.html[Logstash's JVM Settings]. Consider setting it to half of the heap size. Setting too low direct memory decreases the performance of ingestion.
```
karenzone, line 97:
```suggestion
You can set the amount of direct memory with `-XX:MaxDirectMemorySize` in {logstash-ref}/jvm-settings.html[Logstash JVM Settings]. 
Consider setting direct memory to half of the heap size. 
Setting direct memory too low decreases the performance of ingestion.
```

docs/index.asciidoc
```diff
@@ -88,6 +88,17 @@ will be similar to events directly indexed by {plugin-uc} into Elasticsearch.
...
+
+NOTE: Be sure that heap and direct memory combined does not exceed the total memory available on the server to avoid OutOfDirectMemoryError
```
karenzone, line 99: What about "... to avoid an OutOfDirectMemoryError." (singular) or "... to avoid errors."  (plural)

karenzone (approved):
Left some minor suggestions inline.  Otherwise, LGTM

PR #455, opened 2022-08-25
Stop packaging jackson dependencies

robbavey:
Use compileOnly and testImplemenation dependency options to rely on jars
provided by Logstash


andsel (approved):
LGTM
the :red_circle: is for docker image of `7.17.7`not yet available

PR #459, opened 2022-11-11
Identification algorithm

### LINKED ISSUE #458: [stub] add advisory hints to ssl engine to fast-track negotiation
This is a stub issue for tracking to be followed-up with a PR that describes the improvement more succinctly with code than I can do in a formal issue description.

yaauie:
Resolves #458


yaauie:
It turns out my expectations were not quite right, and this hint should be provided with an opt-in flag.

I will come back to it shortly.

PR #461, opened 2022-12-10
Update the `executor_thread` explanation to clarify the default value.

mashhurs:
Clarifies the default value of `executor_thread`.

 - Closes: #460

Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

PR #462, opened 2022-12-13
fix jruby bundler for test

kaisecheng:
Fix: https://github.com/elastic/logstash/issues/14800

Use the bundler in /vendor/bundle/jruby/2.6.0/bin/bundle instead of /vendor/jruby/bin/bundle to run test, as Logstash doesn't package the latter in artifacts in https://github.com/elastic/logstash/pull/14667

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -5,7 +5,7 @@ env
 
 set -ex
 
-jruby -rbundler/setup -S rspec -fd
+bundle exec rspec --format=documentation
 
-jruby -rbundler/setup -S rake test:integration:setup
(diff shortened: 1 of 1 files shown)


.ci/run.sh
```diff
@@ -5,7 +5,7 @@ env
 
 set -ex
 
-jruby -rbundler/setup -S rspec -fd
+bundle exec rspec --format=documentation
```
mashhurs, line 8: Oops! I also added this with my ongoing [PR](https://github.com/logstash-plugins/logstash-input-beats/pull/461).

kaisecheng: ok. I close this one

PR #463, opened 2022-12-14
Add one executor thread per CPU context to the doc.

mashhurs:
This PR intends to cover what I forgot the [last PR suggestion](https://github.com/logstash-plugins/logstash-input-beats/pull/461#discussion_r1047909783) to apply.


andsel (approved):
LGTM

PR #464, opened 2022-12-22
Introduce enrich options to handle ECS passthrough.

mashhurs:
TODOs:
[ ] update/add unit/integration tests
[ ] update docs

### DIFF
--- a/lib/logstash/inputs/beats.rb
+++ b/lib/logstash/inputs/beats.rb
@@ -104,9 +104,9 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   # Enables storing client certificate information in event's metadata. You need 
   # to configure the `ssl_verify_mode` to `peer` or `force_peer` to enable this.
-  config :ssl_peer_metadata, :validate => :boolean, :default => false
+  config :ssl_peer_metadata, :validate => :boolean, :default => false, :deprecated => 'This option will be removed in the future, please add `ssl_peer_metadata` to `enrich` config to use.'
 
-  config :include_codec_tag, :validate => :boolean, :default => true
+  config :include_codec_tag, :validate => :boolean, :default => true, :deprecated => 'This option will be removed in the future, please add `include_codec_tag` to `enrich` config to use'
 
   # Time in milliseconds for an incomplete ssl handshake to timeout
   config :ssl_handshake_timeout, :validate => :number, :default => 10000
@@ -135,6 +135,12 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   # The maximum TLS version allowed for the encrypted connections. The value must be the one of the following:
   # 1.0 for TLS 1.0, 1.1 for TLS 1.1, 1.2 for TLS 1.2
   config :tls_max_version, :validate => :number, :default => TLS.max.version, :deprecated => "Set 'ssl_supported_protocols' instead."
+
+  # The fields to enrich
+  # available modes are `none`, `default` and `all`
+  # default: ['ssl_peer_metadata', 'include_codec_tag']
+  # all: defaults + ['ssl_peer_metadata', 'include_codec_tag']
+  config :enrich, :validate => :array, :default => ['default']
 
   attr_reader :field_hostname, :field_hostip
   attr_reader :field_tls_protocol_version, :field_tls_peer_subject, :field_tls_cipher
@@ -187,6 +193,14 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
     else
       @logger.warn("configured ssl_certificate => #{@ssl_certificate.inspect} will not be used") if @ssl_certificate
       @logger.warn("configured ssl_key => #{@ssl_key.inspect} will not be used") if @ssl_key
+    end
+
+    resolve_enriches
+
+    # Prevent ingest pipelines having an issue when using ECS compatibility where replaces message to event.original field
+    # GH issue: https://github.com/logstash-plugins/logstash-input-elastic_agent/issues/3
+    if !@enrich.include?('event_original') && !@codec.ecs_compatibility.eql?(:disabled) && @codec.class.eql?(LogStash::Codecs::Plain)
+      @codec = LogStash::Codecs::Plain.new('charset' => @codec.charset, 'format' => @codec.format, 'ecs_compatibility' => 'disabled')
     end
 
     # Logstash 6.x breaking change (introduced with 4.0.0 of this gem)
@@ -246,7 +260,7 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
   end
 
   def client_authentication_metadata?
-    @ssl_peer_metadata && ssl_configured? && client_authentification? 
+    @enrich.include?('ssl_peer_metadata') && ssl_configured? && client_authentification?
   end
 
   def client_authentication_required?
@@ -303,4 +317,20 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
     error_details
   end
 
+  def resolve_enriches
+    default_enriches = [ 'event_original', 'source_metadata', 'codec_metadata' ]
+    all_enriches = default_enriches + ['ssl_peer_metadata', 'include_codec_tag']
+
+    @enrich = [] if @enrich.length == 1 && @enrich.first == 'none'
+    @enrich = all_enriches if @enrich.length == 1 && @enrich.first == 'all'
+
+    if @enrich.include?('default')
+      @enrich.delete('default')
+      @enrich |= default_enriches # unique values only
+    end
+
+    @enrich.each do |enrich|
+      raise LogStash::ConfigurationError, "#{enrich} enrich is not supported, please remove." unless all_enriches.include?(enrich)
+    end
+  end
 end
--- a/lib/logstash/inputs/beats/event_transform_common.rb
+++ b/lib/logstash/inputs/beats/event_transform_common.rb
@@ -47,7 +47,7 @@ module LogStash module Inputs class Beats
     end
 
     def include_codec_tag?
-      @input.include_codec_tag
+      @input.enrich.include?('include_codec_tag')
     end
   end
 end; end; end


lib/logstash/inputs/beats.rb
```diff
@@ -104,9 +104,9 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   # Enables storing client certificate information in event's metadata. You need 
   # to configure the `ssl_verify_mode` to `peer` or `force_peer` to enable this.
-  config :ssl_peer_metadata, :validate => :boolean, :default => false
+  config :ssl_peer_metadata, :validate => :boolean, :default => false, :deprecated => 'This option will be removed in the future, please add `ssl_peer_metadata` to `enrich` config to use.'
 
-  config :include_codec_tag, :validate => :boolean, :default => true
+  config :include_codec_tag, :validate => :boolean, :default => true, :deprecated => 'This option will be removed in the future, please add `include_codec_tag` to `enrich` config to use'
```
yaauie, line 109: When a deprecated option is used, the string provided here is an "extra" in a log message that already includes boiler-plate about deprecations and future removal, so we can be more clear with fewer words. The following would encourage our users to look for the docs for how to use the `enrich` option instead of giving them a hint that may have unexpected side-effects:

~~~ ruby
:deprecated => "use `enrich` option to configure which enrichments to perform"
~~~

yaauie (commented):
I have added some suggestions to try to better-communicate the draft-path you are working from. The short version is:
 - even though we are deprecating `ssl_peer_metadata` and `include_codec_tag` options, we still need to support them if they are explicitly provided, and should error if they are provided along-side the new `enrich` option because that would make the config ambiguous.
 - I added better notes about using a "sentinel" codec so that we can change our behaviour _if-and-only-if_ the user didn't explicitly provide us with a codec.
 - I consolidated the originally-specified `event_original` flag for `enrich` into `codec_metadata` since it is metadata about how a codec transforms a sequence of bytes into an event
 - your newly-introduced `include_codec_tag` for `enrich` is also a subset of the as-specified `codec_metadata` since it is metadata about how a codec transforms a sequence of bytes into an event

lib/logstash/inputs/beats.rb
```diff
@@ -303,4 +317,20 @@ def error_details(e, trace = false)
...
+    @enrich = all_enriches if @enrich.length == 1 && @enrich.first == 'all'
+
+    if @enrich.include?('default')
+      @enrich.delete('default')
+      @enrich |= default_enriches # unique values only
+    end
+
+    @enrich.each do |enrich|
+      raise LogStash::ConfigurationError, "#{enrich} enrich is not supported, please remove." unless all_enriches.include?(enrich)
+    end
```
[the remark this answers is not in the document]
mashhurs: Ah thanks. Your proposed alias concept is now clear to me.

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +195,14 @@ def register
       @logger.warn("configured ssl_key => #{@ssl_key.inspect} will not be used") if @ssl_key
     end
 
+    resolve_enriches
+
+    # Prevent ingest pipelines having an issue when using ECS compatibility where replaces message to event.original field
+    # GH issue: https://github.com/logstash-plugins/logstash-input-elastic_agent/issues/3
+    if !@enrich.include?('event_original') && !@codec.ecs_compatibility.eql?(:disabled) && @codec.class.eql?(LogStash::Codecs::Plain)
+      @codec = LogStash::Codecs::Plain.new('charset' => @codec.charset, 'format' => @codec.format, 'ecs_compatibility' => 'disabled')
+    end
```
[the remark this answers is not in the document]
mashhurs: That makes sense to me. 
However, I was failed to set the default sentinel plugin. Current structure seems that DEFAULT (`plain` for the beats-input) plugin has to be initialized. Declaring default codec sentinel will be an issue when pipeline is compiled.

```
[2022-12-27T16:24:58,435][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"Java::JavaLang::IllegalStateException", :message=>"Unable to configure plugins: (NoMethodError) undefined method `id' for #<Object:0x2a076049>", :backtrace=>["org.logstash.config.ir.CompiledPipeline.<init>(CompiledPipeline.java:120)", "org.logstash.execution.AbstractPipelineExt.initialize(AbstractPipelineExt.java:182)",
```
I tried to imitate the methods with followings:

```
DEFAULT_CODEC_SENTINEL.define_singleton_method(:id) do
      "DEFAULT_SENTINEL_CODEC_ID"
    end
    DEFAULT_CODEC_SENTINEL.define_singleton_method(:metric=) do
      Object.new
    end
```

but 1) I don't think for the long term it is a good approach if plugin standard logics will change. 2) it still causes an error by requiring multiple params.

```
[2022-12-27T16:27:22,159][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"Java::JavaLang::IllegalStateException", :message=>"Unable to configure plugins: (ArgumentError) wrong number of arguments (given 1, expected 0)", :backtrace=>["org.logstash.config.ir.CompiledPipeline.<init>(CompiledPipeline.java:120)", "org.logstash.execution.AbstractPipelineExt.initialize(AbstractPipelineExt.java:182)",
```

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +195,14 @@ def register
       @logger.warn("configured ssl_key => #{@ssl_key.inspect} will not be used") if @ssl_key
     end
 
+    resolve_enriches
```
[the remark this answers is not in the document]
mashhurs: In case of naming convention `resolved_enriches`, it is really up to team/developers but I usually use the _past_ tense if I declare static values and make naming _actionable_ ex.: `doSomething/do_something` if method executes some actions. And, in our source base I do see similar approach, ex.: `create_server`. It is minor but sometimes brings some confusions. Let me please know your opinion.

mashhurs:
> I have added some suggestions to try to better-communicate the draft-path you are working from. The short version is:
> 
> * even though we are deprecating `ssl_peer_metadata` and `include_codec_tag` options, we still need to support them if they are explicitly provided, and should error if they are provided along-side the new `enrich` option because that would make the config ambiguous.
> * I added better notes about using a "sentinel" codec so that we can change our behaviour _if-and-only-if_ the user didn't explicitly provide us with a codec.
> * I consolidated the originally-specified `event_original` flag for `enrich` into `codec_metadata` since it is metadata about how a codec transforms a sequence of bytes into an event
> * your newly-introduced `include_codec_tag` for `enrich` is also a subset of the as-specified `codec_metadata` since it is metadata about how a codec transforms a sequence of bytes into an event


It seems `codec_metadata` is a config belong to codec, decides to include/exclude the metadata. **_Does it also control plugin level `metadata` config?_**
 But metadata config is available for some codecs and N/A for some other.
    ex: rubydebug has `metadata` (boolean) config but `plain` codec doesn't have.
    I don't think checking each codec in the source base would be a best practise.

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +206,15 @@ def register
       @logger.warn("configured ssl_key => #{@ssl_key.inspect} will not be used") if @ssl_key
     end
 
+    @enrich = resolve_enriches
+
+    @include_codec_tag = original_params.include?('include_codec_tag') ? params['include_codec_tag'] : include_codec_metadata?
+    @ssl_peer_metadata = original_params.include?('ssl_peer_metadata') ? params['ssl_peer_metadata'] : include_source_metadata?
```
mashhurs, line 212: we are considering `ssl_peer_metadata` in the default (through alias) enriches and applying `source_metadata` makes `ssl_peer_metadata` true if deprecated config is not provided (and we should avoid aligning on deprecated config)

```suggestion
    @ssl_peer_metadata = original_params.include?('ssl_peer_metadata') ? params['ssl_peer_metadata'] : @enrich.include?('ssl_peer_metadata')
```

lib/logstash/inputs/beats.rb
```diff
@@ -58,9 +59,12 @@ class LogStash::Inputs::Beats < LogStash::Inputs::Base
 
   include LogStash::PluginMixins::EventSupport::EventFactoryAdapter
 
+  include LogStash::PluginMixins::PluginFactorySupport
+
   config_name "beats"
 
-  default :codec, "plain"
+  DEFAULT_CODEC_SENTINEL = Object.new
+  default :codec, DEFAULT_CODEC_SENTINEL
```
mashhurs, line 67: Applying default codec sentinel crashes the plugin initialization in the core. [Provided details with the comment](https://github.com/logstash-plugins/logstash-input-beats/pull/464#discussion_r1057624304).

mashhurs: Maybe possible solution align on user's codec and enrich configs as tried with [this commit](https://github.com/logstash-plugins/logstash-input-beats/pull/464/commits/61cdf2204750d23671c0dbb663467abec017f85e) but extra codec initialization will happen if logic needs to re-initialize the codec.

mashhurs:
Open questions from above comments:
1.  Does `codec_metadata` also control plugin level metadata config? As I checked some of plugins such as [`rubydebug` has a separate `metadata` config](https://github.com/logstash-plugins/logstash-codec-rubydebug/blob/main/lib/logstash/codecs/rubydebug.rb#L11) and other plugins don't have it.  
2. How can we [overcome the issue when sentinel codec is initialized](https://github.com/logstash-plugins/logstash-input-beats/pull/464#discussion_r1057624304) (eliminates unnecessary default codec initialization)

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +205,17 @@ def register
       @logger.warn("configured ssl_key => #{@ssl_key.inspect} will not be used") if @ssl_key
     end
 
+    @enrich = resolve_enriches
+
+    @include_codec_tag = original_params.include?('include_codec_tag') ? params['include_codec_tag'] : include_codec_metadata?
+    @ssl_peer_metadata = original_params.include?('ssl_peer_metadata') ? params['ssl_peer_metadata'] : @enrich.include?('ssl_peer_metadata')
+
```
mashhurs, line 212: It seems we may need to warn the user when `enrich` is alone defined but codec is missed.

```suggestion
if !original_params.include?('codec') && @enrich && @enrich.include?('codec_metadata')
      @logger.warn("`enrich` configuration is ignored since `codec` isn't provided. You can add codec definition in the setting to include codec metadata.")
    end
```

lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -32,11 +32,14 @@ def initialize(queue, input)
...
-      unless ip_address.nil? || hash['@metadata'].nil?
-        set_nested(hash, @input.field_hostip, ip_address)
+      if @input.include_source_metadata?
+        ip_address = ip_address(ctx)
+        unless ip_address.nil? || hash['@metadata'].nil?
+          set_nested(hash, @input.field_hostip, ip_address)
+        end
       end
+
       target_field = extract_target_field(hash)
```
mashhurs, line 43: Why do we really need to override the codec? if we allow raw transfer when customer doesn't want codec metadata, it seems the problem will be solved!

```suggestion
      target_field = @input.include_codec_metadata?? extract_target_field(hash) : nil
```

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +205,17 @@ def register
...
+    # intentionally ask users to provide codec when they want to use the codec metadata
+    # second layer enrich is also a controller, provide enrich => ['codec_metadata' or 'default'] with codec if you override event original
+    if !original_params.include?('codec') || (@enrich && (!@enrich.include?('codec_metadata') || @enrich.include?('none')))
```
mashhurs, line 215: Self note: this is a breaking change! Clarify with @yaauie and suggest raw transformation (`lib/logstash/inputs/beats/message_listener.rb:L43`)

docs/index.asciidoc
```diff
@@ -415,6 +420,35 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+|=======================================================================
+|Option |Description
+
+|`none`                  |Not to include any data
+|`default` alias | `codec_metadata` and `codec_metadata` will be included
+|`all` alias | All `codec_metadata`, `codec_metadata` and `ssl_peer_metadata` will be included
+|`source_metadata` | Source metadata such as host name will be included
+|`codec_metadata` | Codec metadata such as `tag`, `[event][original]` will be included
+|`ssl_peer_metadata` | SSL peer metadata will be included
+|=======================================================================
```
yaauie, line 440: There is some nuance here with how `enrich` interacts with the ECS Compatibility fields table at the top of this doc that call out "availability" as "always" etc. I'll take a stab at re-tabling that info separately, but below we have an in-list table (asciidoc: bare `+` on a newline causes the following block element to be injected to the current list item)

```suggestion
[id="plugins-{type}s-{plugin}-enrich"]
===== `enrich`

  * Value type is <<string,string>>
  ** A <<list,list>> can also be provided
  ** Configures which enrichments are applied to each event
  ** Supported values are:
+
[cols="2l,5"]
|=======================================================================
|Enrichment | Description

| codec_metadata    | Information about how the codec transformed a sequence of bytes into
                      this Event, such as _which_ codec was used. If no <<codec>> is explicitly
                      specified, _excluding_ `codec_metadata` from `enrich` will prevent the
                      default codec itself from adding metadata such as `[event][original]`.
| source_metadata   | Information about the _source_ of the event, such as the IP address
                      of the inbound connection this input received the event from and the
                      name of the Logstash host that processed the event
| ssl_peer_metadata | Detailed information about the _SSL peer_ we received the event from,
                      such as identity information from the SSL client certificate that was
                      presented when establishing a connection to this input
| default           | _alias_ resolving to _at least_ `[codec_metadata, source_metadata]`
| all               | _alias_ to include _all_ available enrichments (including additional
                      enrichments introduced in future versions of this plugin)
| none              | _alias_ to _exclude_ all available enrichments
|=======================================================================
  ** Default value is `default`, which is an alias for for `[codec_metadata, source_metadata]` that may be extended in future versions of this plugin to include additional by-default enrichments.

```

docs/index.asciidoc
```diff
@@ -415,6 +420,35 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+By default, the plugin initializes ECS-disabled plain codec, source and codec metadata such as source host or `[event][original]` will not be added.
+`enrich` option helps to manage which metadata to be included. Available options are:
+
+[cols="<,<"]
+|=======================================================================
+|Option |Description
+
+|`none`                  |Not to include any data
+|`default` alias | `codec_metadata` and `codec_metadata` will be included
+|`all` alias | All `codec_metadata`, `codec_metadata` and `ssl_peer_metadata` will be included
```
roaksoax, line 436: Do we really need the `default` or `all` aliases? I would argue that we should completely drop them. So that:

1. The documentation should highlight that the default behavior (when the `enrich` option is NOT provided) is to default to `codec_metadata` and `ssl_peer_metadata`.
2. If the user wants to disable enrichment OR override the default behavior, then they can add it.

As such, having a `default` option makes no sense to me because not providing the option is already default. It would just be repetitive and IMHO, useless.


As far as the `all` alias, while I understand that we would be making it easier for users to benefit from all the metadata additions, it would also potentially be a reason why customers change the settings after upgrade if we add a new metadata option. Instead, users should be able to *opt-in* into that option if they have manually changed the `enrich` option.

roaksoax (changes requested):
@mashhurs @yaauie Hi folks, I know I'm late to the party, but I have some feedback that should be considered before calling this done. Please see inline.

I'll block this with "Request changes" just in case.

docs/index.asciidoc
roaksoax: Also, it is fine to drop the "default" alias, e.g., it doesn't make sense to do:

enrich => ["default"]

roaksoax: After chatting with @yaauie, and going over the PR the solution for this is to:

1. Make documentation that clearly describes what the default behavior is *when not providing the option*
2. The `default` option, should be an internal thing, not something it makes sense for users to set (e.g. enrich => default). It brings no value to have that as an option users can change. To simplify, we should only make doc changes to accordingly describe the default behavior.
3. I'm fine keeping `all` as a user option.

lib/logstash/inputs/beats/message_listener.rb
mashhurs: After stepping back and thinking, I think we are giving more flexibility for the users to control the metadatas rather than switching off (aligning on raw transformer) the codec.

docs/index.asciidoc
```diff
@@ -415,6 +420,35 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+|`codec_metadata` | Codec metadata such as `tag`, `[event][original]` will be included
+|`ssl_peer_metadata` | SSL peer metadata will be included
+|=======================================================================
+
+As an example, following configuration enables source and metadata with ECS enabled plain codec:
+```
+input {
+  beats {
+    port => 5044
+    enrich => ["default"]
```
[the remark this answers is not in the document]
mashhurs: Applied the change and now `enrich` can be provided as a you mentioned string or list of strings.

docs/index.asciidoc
mashhurs: Thanks for the feedback @roaksoax and it makes sense and I have updated the logic.

docs/index.asciidoc
```diff
@@ -415,6 +420,35 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
 The minimum TLS version allowed for the encrypted connections.
 The value must be one of the following: 1.1 for TLS 1.1, 1.2 for TLS 1.2, 1.3 for TLS 1.3
 
+[id="plugins-{type}s-{plugin}-enrich"]
+===== `enrich`
```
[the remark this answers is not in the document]
mashhurs: We [tried to design the config](https://github.com/logstash-plugins/logstash-input-elastic_agent/issues/3#issuecomment-1343231583) to be "open to extend" in the future. I am okay to rename and give narrowband but one thing if we know the non-metadata enrichment use case, we would have more ideas how to optimally design the flow.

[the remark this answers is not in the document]
roaksoax: ack! I'm fine using the naming.

lib/logstash/inputs/beats/message_listener.rb
yaauie: I understand your "multi-depth becomes the 1-depth" to be only a problem if:
(a) the structured event contains a top-level `message` _field_ with a _string_ value that holds a JSON-encoded payload
(b) the user instantiates this plugin with `codec => json`, causing the _decoded_ payload to get merged into the event as-sent by Beats

Because our _default_ codec is `plain`, and it creates an event from the `message` and the result is an event with a `message` that gets merged on-top of the existing `message` field, I don't see this to be practically a problem.

However, what you link to is likely a bug in which the `[event][original]` produced by the codec could _override_ an `[event][original]` that is part of the original beats payload :weary:

spec/inputs/beats_spec.rb
```diff
@@ -200,6 +200,78 @@
...
+    context "with default codec configuration" do
+      it "should load default ['source_metadata', 'codec_metadata'] configs" do
+        plugin = LogStash::Inputs::Beats.new(config)
+        plugin.register
+        expect(plugin.include_codec_tag).to be_truthy
+        expect(plugin.ssl_peer_metadata).to be_falsey
+
+        es_major_version = SpecHelper.es_major_version.to_i
+        expect(plugin.codec.ecs_compatibility).to eq(es_major_version > 7 ? :v8 : :disabled)
+      end
```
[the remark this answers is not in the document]
yaauie: _[ EDIT: I ended up fleshing this out in mashhurs/logstash-input-beats#2 ]_

If we were to define some shared examples, we could make simple use of them:

~~~ ruby
example_group "source_metadata enabled" do
  it "is configured to enrich source_metadata" do
    expect(registered_plugin.enrich).to include("source_metadata")
  end
end

example_group "source_metadata disabled" do
  it "is configured to not enrich source_metadata" do
    expect(registered_plugin.enrich).to_not include("source_metadata")
  end
end

example_group "codec_metadata enabled" do
  it "is configured to enrich `codec_metadata`"
    expect(registered_plugin.enrich).to include("codec_metadata")
    expect(registered_plugin.include_codec_tag).to be_truthy

    expect(registered_plugin.codec.original_params).to be_empty
  end
end

example_group "codec_metadata disabled" do
  it "is configured to not enrich `codec_metadata`"
    expect(registered_plugin.enrich).to_not include("codec_metadata")
    expect(registered_plugin.include_codec_tag).to be_falsey

    # note: disabling ECS is an _implementation detail_ of how we prevent
    # the codec from enriching the event with [event][original]
    expect(registered_plugin.codec.original_params).to include('ecs_compatibility' => 'disabled')
  end

  context "and a codec is explicitly provided" do
    let(:config) { super().merge("codec" => "plain") }
    it "does not override the codec" do

      # note: disabling ECS is an _implementation detail_ of how we prevent
      # the codec from enriching the event with [event][original], so we ensure
      # the absence of the setting.
      expect(registered_plugin.codec.original_params).to_not include('ecs_compatibility')
    end
  end
end

example_group "ssl_peer_metadata enabled" do
  it "is configured to enrich ssl_peer_metadata" do
    expect(registered_plugin.enrich).to include("ssl_peer_metadata")
    expect(registered_plugin.ssl_peer_metadata).to be_truthy
  end
end

example_group "ssl_peer_metadata disabled" do
  it "is configured to not enrich ssl_peer_metadata" do
    expect(registered_plugin.enrich).to_not include("ssl_peer_metadata")
    expect(registered_plugin.ssl_peer_metadata).to be_falsey
  end
end
~~~


~~~ ruby
context "enrich" do
  let(:plugin) { LogStash::Inputs::Beats.new(config) }
  let(:registered_plugin) { plugin.tap(&:register) }
  context "defaults" do
    include_examples "codec_metadata enabled"
    include_examples "source_metadata enabled"
    include_examples "ssl_peer_metadata disabled"
  end
  context "all" do
    let(:config) { super().merge("enrich" => "all") }
    include_examples "codec_metadata enabled"
    include_examples "source_metadata enabled"
    include_examples "ssl_peer_metadata enabled"
  end
  context "none" do
    let(:config) { super().merge("enrich" => "none") }
    include_examples "codec_metadata disabled"
    include_examples "source_metadata disabled"
    include_examples "ssl_peer_metadata disabled"
  end
  context "empty list" do
    let(:config) { super().merge("enrich" => []) }
    include_examples "codec_metadata disabled"
    include_examples "source_metadata disabled"
    include_examples "ssl_peer_metadata disabled"
  end

  %w(
    codec_metadata
    source_metadata
    ssl_peer_metadata
  ).each do |single_active_enrichment|
    context "only #{single_active_enrichment} enabled" do
      let(:config) { super().merge("enrich" => single_active_enrichment) }
      include_examples "codec_metadata #{single_active_enrichment == "codec_metadata" ? "enabled" : "disabled"}"
      include_examples "source_metadata #{single_active_enrichment == "source_metadata" ? "enabled" : "disabled"}"
      include_examples "ssl_peer_metadata #{single_active_enrichment == "ssl_peer_metadata" ? "enabled" : "disabled"}"
    end
  end
end
~~~

lib/logstash/inputs/beats.rb
```diff
@@ -257,6 +288,14 @@ def require_certificate_authorities?
...
   end
 
+  def include_source_metadata?
+    return @enrich.include?('source_metadata')
+  end
+
+  def include_codec_metadata?
+    return @enrich.include?('codec_metadata')
+  end
+
```
yaauie, line 298: While `include_codec_metadata?` is _only_ used in `register` to effectively set `@include_codec_tag` (which is the source-of-truth for that action), and only does so when the deprecated `include_codec_tag` is not provided, its presence means that we publicly provide a method that doesn't tell the whole story.

We also should avoid reaching into `@enrich` for every single event since this is known-constant when the plugin is registered.

I've opened https://github.com/mashhurs/logstash-input-beats/pull/2 to refactor `register` to do _all_ of the work, using the alias-resolved enrich as a local variable so that we (a) don't leave traps for our future selves and (b) don't add per-event overhead.

yaauie (commented):
Actual functionality looks good.

I've opened two PR's against your branch:

 - [one](https://github.com/mashhurs/logstash-input-beats/pull/1/files) refactors the existing docs for ECS and metadata fields, since the introduction of `enrich` invalidates much of what it communicates
 - [two](https://github.com/mashhurs/logstash-input-beats/pull/2/files) refactors the use of the resolved enrichments to _populate_ the source-of-truth for each enrichment instead of this PR's mixed sources-of-truth, so that we don't need to query that resolved array for each event. This PR also refactors the specs to make extensive use of shared examples so that we can validate that each of our aliases and ways of specifying the enrichments works as-intended. These examples specify _behaviour_ and call out which implementation details it is using to validate the behaviour _so that we can change those details at a later date without breaking our specification_.

spec/spec_helper.rb
```diff
@@ -13,3 +13,25 @@
...
+module SpecHelper
+
+  def self.es_major_version
+    es_version.split('.').first
+  end
+
+  def self.es_version
+    [
+      nilify(RSpec.configuration.filter[:es_version]),
+      nilify(ENV['ES_VERSION']),
```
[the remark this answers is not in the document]
mashhurs: My focus point was codec's behavior which re-shapes the origin data and coming back to enrichment state (based on alias & available configs) make sense.
Thanks for clarifying this.

docs/index.asciidoc
```diff
@@ -415,6 +459,60 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+
+  * Value type is <<string,string>>
+  ** A <<list,list>> can also be provided
+  ** Configures which enrichments are applied to each event
+  ** Supported values are:
++
+[cols="2l,5"]
+|=======================================================================
+|Enrichment | Description
```
mashhurs, line 472: @karenzone I would appreciate if you take a quick look at the changes. Thanks.

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
-|ECS `disabled` |ECS `v1`, `v8` |Availability |Description
-
-|[host]                  |[@metadata][input][beats][host][name]   |Always       |Name or address of the {plugin-singular} host
-|[@metadata][ip_address] |[@metadata][input][beats][host][ip]     |Always       |IP address of the {plugin-uc} client
-|[@metadata][tls_peer][status] | [@metadata][tls_peer][status] | When SSL related fields are populated | Contains "verified"/"unverified" labels in `disabled`, `true`/`false` in `v1`/`v8`
-|[@metadata][tls_peer][protocol] | [@metadata][input][beats][tls][version_protocol] | When SSL status is "verified" | Contains the TLS version used (e.g. `TLSv1.2`)
-|[@metadata][tls_peer][subject] | [@metadata][input][beats][tls][client][subject] | When SSL status is "verified" | Contains the identity name of the remote end (e.g. `CN=artifacts-no-kpi.elastic.co`)
-|[@metadata][tls_peer][cipher_suite] | [@metadata][input][beats][tls][cipher] | When SSL status is "verified" | Contains the name of cipher suite used (e.g. `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`)
+
+.`ssl_peer_metadata``
```
mashhurs, line 171: Seems extra back tick we have
```suggestion
.`ssl_peer_metadata`
```

yaauie:
Since this is now a combined work with my own commits on top, it is unlikely that further review from me will turn up anything that I missed. @jsvd has agreed to lend his fresh eyes to for the final review.

jsvd:
A few tests:

###  Event with "message" field

`{ "message" => "event with message 完璧"}`

#### Logstash Without this PR `beats { codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:49:22.936160Z,
       "message" => "event with message å®\u008Cç\u0092§",
         "event" => {
        "original" => "event with message å®\u008Cç\u0092§"
    },
          "tags" => [
        [0] "beats_input_codec_plain_applied"
    ]
}
```

#### Logstash with this PR and `beats { enrich => all codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:50:24.807838Z,
         "event" => {
        "original" => "event with message å®\u008Cç\u0092§"
    },
       "message" => "event with message å®\u008Cç\u0092§",
          "tags" => [
        [0] "beats_input_codec_plain_applied"
    ]
}
```

#### Logstash with this PR and `beats { enrich => none codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:50:25.697849Z,
         "event" => {
        "original" => "event with message å®\u008Cç\u0092§"
    },
       "message" => "event with message å®\u008Cç\u0092§"
}

```

#### Logstash with this PR and `beats { enrich => none }`

```
{
      "@version" => "1",
       "message" => "event with message 完璧",
    "@timestamp" => 2023-01-25T11:56:19.964199Z
}
```

###  Event with nested event.original field

`{ "event" => { "original" => "nested event -> original field 完璧"} }`

#### Logstash Without this PR `beats { codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:49:22.946685Z,
       "message" => "original message å®\u008Cç\u0092§",
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
          "tags" => [
        [0] "beats_input_codec_plain_applied"
    ]
}
```
#### Logstash with this PR and `beats { enrich => all codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:50:24.810634Z,
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
          "tags" => [
        [0] "beats_input_raw_event"
    ]
}
```

#### Logstash with this PR and `beats { enrich => none codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:50:25.698132Z,
         "event" => {
        "original" => "nested event -> original field 完璧"
    }
}
```

#### Logstash with this PR and `beats { enrich => none }`

```
{
      "@version" => "1",
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
    "@timestamp" => 2023-01-25T11:56:19.966700Z
}
```

###  Event with nested event.original field and message field

`{ "event" => { "original" => "nested event -> original field 完璧"}, "message": "original message 完璧"}`

#### Logstash Without this PR `beats { codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:49:22.945690Z,
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
          "tags" => [
        [0] "beats_input_raw_event"
    ]
}

```
#### Logstash with this PR and `beats { enrich => all codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:50:24.810925Z,
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
       "message" => "original message å®\u008Cç\u0092§",
          "tags" => [
        [0] "beats_input_codec_plain_applied"
    ]
}
```

#### Logstash with this PR and `beats { enrich => none codec => plain { charset => 'ISO8859-1' } }`

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:50:25.698278Z,
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
       "message" => "original message å®\u008Cç\u0092§"
}
```

#### Logstash with this PR and `beats { enrich => none }`

```
{
      "@version" => "1",
         "event" => {
        "original" => "nested event -> original field 完璧"
    },
       "message" => "original message 完璧",
    "@timestamp" => 2023-01-25T11:56:19.966879Z
}
```

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +206,22 @@ def register
...
+    @include_source_metadata = active_enrichments.include?('source_metadata')
+    @include_codec_tag = original_params.include?('include_codec_tag') ? params['include_codec_tag'] : active_enrichments.include?('codec_metadata')
+    @ssl_peer_metadata = original_params.include?('ssl_peer_metadata') ? params['ssl_peer_metadata'] : active_enrichments.include?('ssl_peer_metadata')
+
+    # intentionally ask users to provide codec when they want to use the codec metadata
+    # second layer enrich is also a controller, provide enrich => ['codec_metadata' or/with 'source_metadata'] with codec if you override event original
+    unless active_enrichments.include?('codec_metadata')
+      if original_params.include?('codec')
+        @logger.warn('`enrich` configuration does not include `codec_metadata`, but the directive is not propagated to the explicitly-defined `codec`')
```
jsvd, line 219: As a non native English speaker it's hard for me gauge the consequences of this warning. In fact, getting this warning after setting "codec_metadata" in enrich w/ an explicit codec suggests we won't be skipping the  `beats_input_codec_plain_applied`/`beats_input_raw_event` tags, but we do (see `enrich => none codec => plain { charset => 'ISO8859-1' } }` example on my big comment). However we don't warn about not disabling ECS compatibility anymore.

I believe we should have a clearer relationship between ECS compatibility and the codec_metadata enrich option.
I see a few alternatives:
1. creating a separate enrich option just for the ECS toggling
2. expanding codec_metadata to be....`codec_metadata_and_ecs`?

mashhurs: Warning message: `the directive is not propagated` statement would be a reason for misunderstanding (or not understanding). Paraphrasing with some options where we can choose one of them or give your options please:
- "`enrich` configuration does not include `codec_metadata`. Explicitly defined `codec` will apply ECS compatibility by default or align on `codec` `ecs_compatibility` setting."
- "`codec` is specified but `enrich` configuration does not include `codec_metadata`. ECS compatibility is enabled by default or align on `codec` `ecs_compatibility` configuration."
- "`codec_metadata` option is not include in `enrich` configuration. This will lead the `codec` to control ECS compatibility (by default enabled) based on its `ecs_compatibility` configuration."

1. As my understanding, we have `ecs_compatibility` which gives full control codec's ECS behavior and we have now `codec_metadata` to set the codec behavior. I need more information about the enrich option you are mentioning and how it should work, its relations with currently available settings (such as `ecs_compatibility`), etc...
2. `codec_metadata` -> `codec_metadata_and_ecs`
 (I understood expanding codec_metadata to be `codec_metadata_and_ecs` as a naming change - plz correct me if its meaning is different)
It is true the currently we are mostly referring to `[event][original]` ECS behavior case but `codec_metadata`, a broader naming would give us extra room to control metadata cases in the future.

docs/index.asciidoc
```diff
@@ -415,6 +459,60 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+  ** Supported values are:
++
+[cols="2l,5"]
+|=======================================================================
+|Enrichment | Description
+
+| codec_metadata    | Information about how the codec transformed a sequence of bytes into
+                      this Event, such as _which_ codec was used. If no <<codec>> is explicitly
+                      specified, _excluding_ `codec_metadata` from `enrich` will prevent the
+                      default codec itself from adding metadata such as `[event][original]`.
```
jsvd, line 477:
```suggestion
                      this Event, such as _which_ codec was used. Also, if no <<codec>> is explicitly
                      specified, _excluding_ `codec_metadata` from `enrich` will disable `ecs_compatibility` for this plugin.
```

lib/logstash/inputs/beats.rb
jsvd: I think there is a challenge in explaining the interaction between `enrich => codec_metadata` and `codec`:

If no `codec` it specified and:
a) `enrich => codec_metadata`: plugin adds codec tags and does not influence `ecs_compatibility`
b) `enrich => none`: plugin does not add codec tags and does not influence `ecs_compatibility`

If `codec` is not specified and:
a) `enrich => codec_metadata`: plugin adds codec tags does not influence `ecs_compatibility`
a) `enrich => none`: plugin does not add codec tags **and disables `ecs_compatibility`**

Two potential surprises when `ecs_compatibility` isn't influenced:
1) it's possible for `beats { port => 3333 enrich => codec_metadata }` to not add [event][original], if `pipeline.ecs_compatibility: disabled` is configured in the logstash.yml.
2) Contrary, it's possible for `enrich => none` to still add [event][original] if a codec is set.

I don't think there's a single great option here, unfortunately. The disabling of ECS compatibility through enrich is a very useful convenience but it doesn't fit 100% into the logic of the settings we provide. The next best thing really comes down to providing enough documentation and logging of setting behaviour.
Maybe in the future we can just agree that `ecs_compatibility` does not make sense for this plugin.

Until then we can go with one of the messages you described. Maybe a little tweak:

> "codec is specified but enrich configuration does not include codec_metadata. ECS compatibility is enabled by default or align on codec ecs_compatibility configuration."

We can also add a debug logging entry for the other branch of the conditional:

> "disabling 'ecs_compatibilty' for the default codec since 'enrich' configuration does not include 'codec_metadata' and no explicit codec is set"

mashhurs: I think you mean by `If _no_ codec it specified and:` is `If codec is specified and:`, right? 

I have slightly improved the doc and the logger messages with [this commit](https://github.com/logstash-plugins/logstash-input-beats/pull/464/commits/548e99e2cd58ee80beebd0d5bca52d6a27ef0fae) as your per request and I TOTALLY AGREE with the opinion of **_Maybe in the future we can just agree that `ecs_compatibility` does not make sense for this plugin_** if we think from Elastic stack (+beats, agent) point of view. That was the reason why I initially [pushed _no-transformation_ idea](https://github.com/logstash-plugins/logstash-input-beats/pull/464#discussion_r1064348212). However, this would be a breaking change.

lib/logstash/inputs/beats.rb
```diff
@@ -189,6 +206,23 @@ def register
...
+
+    @include_source_metadata = active_enrichments.include?('source_metadata')
+    @include_codec_tag = original_params.include?('include_codec_tag') ? params['include_codec_tag'] : active_enrichments.include?('codec_metadata')
+    @ssl_peer_metadata = original_params.include?('ssl_peer_metadata') ? params['ssl_peer_metadata'] : active_enrichments.include?('ssl_peer_metadata')
+
+    # intentionally ask users to provide codec when they want to use the codec metadata
+    # second layer enrich is also a controller, provide enrich => ['codec_metadata' or/with 'source_metadata'] with codec if you override event original
+    unless active_enrichments.include?('codec_metadata')
+      if original_params.include?('codec')
+        @logger.warn('`codec` is specified but enrich configuration does not include `codec_metadata`. ECS compatibility aligns on pipeline or codec `ecs_compatibility` configuration, enabled by default.')
```
jsvd, line 219:
```suggestion
        @logger.warn('An explicit `codec` is specified but `enrich` does not include `codec_metadata`. ECS compatibility will remain aligned on the pipeline or codec's `ecs_compatibility` (enabled by default).')
```

lib/logstash/inputs/beats.rb
jsvd: > I think you mean by If _no_ codec it specified and: is If codec is specified and:, right?

correct. fixed.

> That was the reason why I initially https://github.com/logstash-plugins/logstash-input-beats/pull/464#discussion_r1064348212. However, this would be a breaking change.

Indeed, small steps. with `enrich` we have a door towards influencing event manipulation. And we can perhaps decide to deprecate ecs_compatibilty alongside other settings like the ssl and codec enrichments.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 6.5.0
+ - An enrichment `enrich` option added to control ECS passthrough. `ssl_peer_metadata` and `include_codec_tag` configurations are deprecated and can be managed through the `enrich`  [#464](https://github.com/logstash-plugins/logstash-input-beats/pull/464)
```
mashhurs, line 2: @karenzone we need a doc review to release this change. Thank you!

jsvd:
@ruflin FYI, this is what we're going to ship for Logstash to address the "passthrough" need we discussed in person: adding an `enrich` setting that performs some enrichments by default, which can be either disabled one by one or with shortcuts like "none" or "all".

For example, if Agent/Beats sends a document with this structure:

`{ "@timestamp": "2023-01-25T11:49:22.936160Z", "message" => "event with message 完璧"}`

Logstash with/without this PR with the standard default configuration `input { beats { } }` results in this Logstash event:

```
{
      "@version" => "1",
    "@timestamp" => 2023-01-25T11:49:22.936160Z,
       "message" => "event with message 完璧",
         "event" => {
        "original" => "event with message 完璧"
    },
          "tags" => [
        [0] "beats_input_codec_plain_applied"
    ]
}
```

Logstash with this PR and `input { beats { enrich => none } }` results in this Logstash event:

```
{
      "@version" => "1",
       "message" => "event with message 完璧",
    "@timestamp" => 2023-01-25T11:49:22.936160Z
}
```

Other examples above https://github.com/logstash-plugins/logstash-input-beats/pull/464#issuecomment-1403504204

jsvd (approved):
LGTM from functionality and code p.o.v.
Due to heavy changes on documentation please wait for a review from Karen.

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
 
 
 [id="plugins-{type}s-{plugin}-ecs_metadata"]
-==== Event Metadata and the Elastic Common Schema (ECS)
+==== Event Enrichment and the Elastic Common Schema (ECS)
```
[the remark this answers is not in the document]
karenzone: Elastic standard: sentence case for headings.
Headings are inconsistent throughout, but we're trying to correct inconsistencies when we get a chance.

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
 
-When decoding {plugin-uc} events, this plugin adds two fields related to the event:
-the deprecated `host` which contains the `hostname` provided by {plugin-uc} and the
-`ip_address` containing the remote address of the client's connection. When
-<<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is enabled
-these are now moved in ECS compatible namespace. Here's how
-<<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> affects
-output.  
+When decoding {plugin-uc} events, this plugin _enriches_ each event with metadata to make information about where the event came from available during further processing.
```
karenzone, line 148:
```suggestion
When decoding {plugin-uc} events, this plugin enriches each event with metadata about the event's source, making this information available during further processing. 
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+The <<plugins-{type}s-{plugin}-enrich>> option can be used to activate or deactivate individual enrichment categories.
```
karenzone, line 149:
```suggestion
You can use the <<plugins-{type}s-{plugin}-enrich>> option to activate or deactivate individual enrichment categories.
```

karenzone: Use active voice to put the focus back on the user.

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
 
-[cols="<l,<l,e,<e"]
+The location of these enrichment fields depends on whether <<plugins-{type}s-{plugin}-ecs_compatibility,ECS compatibility mode>> is enabled:
+
+- When ECS compatibility is _enabled_, enrichment fields will be added in an ECS-compatible structure.
+- When ECS compatibility is _disabled_, enrichment fields will be added in a way that is backward-compatible with this plugin, but is known to clash with the Elastic Common Schema.
```
karenzone, line 154:
```suggestion
- When ECS compatibility is _enabled_, enrichment fields are added in an ECS-compatible structure.
- When ECS compatibility is _disabled_, enrichment fields are added in a way that is backward-compatible with this plugin, but is known to clash with the Elastic Common Schema.
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+
+
+.`source_metadata`
+[cols="<l,<l,<e",caption="Enrichment Category:"]
```
karenzone, line 158:
```suggestion
[cols="<l,<l,<e",caption="Enrichment category:"]
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+|=======================================================================
+|ECS `v1`, `v8` |ECS `disabled` |Description
+
+|[@metadata][input][beats][host][name]
+|[host]
+|Name or address of the {plugin-singular} host
```
karenzone, line 164: Yay! Bonus points for using the attribute. :-)

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
-
-|[host]                  |[@metadata][input][beats][host][name]   |Always       |Name or address of the {plugin-singular} host
-|[@metadata][ip_address] |[@metadata][input][beats][host][ip]     |Always       |IP address of the {plugin-uc} client
-|[@metadata][tls_peer][status] | [@metadata][tls_peer][status] | When SSL related fields are populated | Contains "verified"/"unverified" labels in `disabled`, `true`/`false` in `v1`/`v8`
-|[@metadata][tls_peer][protocol] | [@metadata][input][beats][tls][version_protocol] | When SSL status is "verified" | Contains the TLS version used (e.g. `TLSv1.2`)
-|[@metadata][tls_peer][subject] | [@metadata][input][beats][tls][client][subject] | When SSL status is "verified" | Contains the identity name of the remote end (e.g. `CN=artifacts-no-kpi.elastic.co`)
-|[@metadata][tls_peer][cipher_suite] | [@metadata][input][beats][tls][cipher] | When SSL status is "verified" | Contains the name of cipher suite used (e.g. `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`)
+
+.`ssl_peer_metadata`
+[cols="<l,<l,<e",caption="Enrichment Category:"]
```
karenzone, line 172:
```suggestion
[cols="<l,<l,<e",caption="Enrichment category:"]
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+|=======================================================================
+|ECS `v1`, `v8` |ECS `disabled` |Description
+
+|[@metadata][tls_peer][status]
+|[@metadata][tls_peer][status]
+|Contains "verified" or "unverified" label; available when SSL is enabled.
+
+|[@metadata][input][beats][tls][version_protocol]
+|[@metadata][tls_peer][protocol]
+|Contains the TLS version used (e.g. `TLSv1.2`); available when SSL status is "verified"
```
karenzone, line 182:
```suggestion
|Contains the TLS version used (such as `TLSv1.2`); available when SSL status is "verified"
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+
+|[@metadata][input][beats][tls][client][subject]
+|[@metadata][tls_peer][subject]
+|Contains the identity name of the remote end (e.g. `CN=artifacts-no-kpi.elastic.co`); available when SSL status is "verified"
```
karenzone, line 186:
```suggestion
|Contains the identity name of the remote end (such as `CN=artifacts-no-kpi.elastic.co`); available when SSL status is "verified"
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+
+|[@metadata][input][beats][tls][cipher]
+|[@metadata][tls_peer][cipher_suite]
+|Contains the name of cipher suite used (e.g. `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`); available when SSL status is "verified"
```
karenzone, line 190:
```suggestion
|Contains the name of cipher suite used (such as `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256`); available when SSL status is "verified"
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+|=======================================================================
+
+.`codec_metadata`
+[cols="<l,<l,<e",caption="Enrichment Category:"]
```
karenzone, line 194:
```suggestion
[cols="<l,<l,<e",caption="Enrichment category:"]
```

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+|=======================================================================
+|ECS `v1`, `v8` |ECS `disabled` |Description
+
+|[tag]
+|[tag]
+|Contains `beats_input_codec_XXX_applied` where `XXX` is the name of the codec
```
karenzone, line 200: Remember that anything you hardcode will show up in the Elastic Agent doc as well.  As this is just an example, consider renaming the example to something generic that would work for both Beats and Agent inputs.

docs/index.asciidoc
```diff
@@ -143,26 +143,65 @@ endif::[]
...
+
+|[event][original]
+e|N/A
+|When ECS is enabled, this plugin's *default codec* will ensure that the event's `[event][original]` field is populated, using the bytes as-processed by the codec if the field does not already exist on the event being processed.
```
karenzone, line 204:
```suggestion
|When ECS is enabled, this plugin's *default codec* ensures that the event's `[event][original]` field is populated. If the field does not already exist on the event being processed, the field uses the bytes as-processed by the codec.
```

karenzone: This sentence was confusing. I made a a guess, but I'm not sure I got it right. 
When sentences are long and convoluted, they become harder to mentally parse and understand. Breaking the content into two or more sentences is a reliable remedy.

docs/index.asciidoc
```diff
@@ -179,19 +218,20 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-ecs_compatibility>> | <<string,string>>|No
 | <<plugins-{type}s-{plugin}-executor_threads>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-host>> |<<string,string>>|No
-| <<plugins-{type}s-{plugin}-include_codec_tag>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-include_codec_tag>> |<<boolean,boolean>>|__Deprecated__
```
karenzone, line 221: More bonus points for correctly notating deprecations!  🎉

docs/index.asciidoc
```diff
@@ -179,19 +218,20 @@ This plugin supports the following configuration options plus the <<plugins-{typ
...
 | <<plugins-{type}s-{plugin}-ssl_handshake_timeout>> |<<number,number>>|No
 | <<plugins-{type}s-{plugin}-ssl_key>> |a valid filesystem path|No
 | <<plugins-{type}s-{plugin}-ssl_key_passphrase>> |<<password,password>>|No
-| <<plugins-{type}s-{plugin}-ssl_peer_metadata>> |<<boolean,boolean>>|No
+| <<plugins-{type}s-{plugin}-ssl_peer_metadata>> |<<boolean,boolean>>|__Deprecated__
 | <<plugins-{type}s-{plugin}-ssl_supported_protocols>> |<<array,array>>|No
 | <<plugins-{type}s-{plugin}-ssl_verify_mode>> |<<string,string>>, one of `["none", "peer", "force_peer"]`|No
 | <<plugins-{type}s-{plugin}-tls_max_version>> |<<number,number>>|__Deprecated__
 | <<plugins-{type}s-{plugin}-tls_min_version>> |<<number,number>>|__Deprecated__
+| <<plugins-{type}s-{plugin}-enrich>> |<<string,string>>|No
```
karenzone, line 234: Should appear alphabetically in both the list table and description.

docs/index.asciidoc
```diff
@@ -415,6 +459,62 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
 The minimum TLS version allowed for the encrypted connections.
 The value must be one of the following: 1.1 for TLS 1.1, 1.2 for TLS 1.2, 1.3 for TLS 1.3
 
+[id="plugins-{type}s-{plugin}-enrich"]
+===== `enrich`
```
karenzone, line 463: Should appear alphabetically in both the list table and description.

docs/index.asciidoc
```diff
@@ -415,6 +459,62 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+| ssl_peer_metadata | Detailed information about the _SSL peer_ we received the event from,
+                      such as identity information from the SSL client certificate that was
+                      presented when establishing a connection to this input
+| all               | _alias_ to include _all_ available enrichments (including additional
+                      enrichments introduced in future versions of this plugin)
+| none              | _alias_ to _exclude_ all available enrichments. Note that, _explicitly_
+                      defining <<codec>> with this option will not disable the `ecs_compatibility`,
+                      instead it relies on pipeline or codec `ecs_compatibility` configuration.
+|=======================================================================
+  ** Default value is `[codec_metadata, source_metadata]` that may be extended in future versions of this plugin to include additional enrichments.
```
karenzone, line 490: Please consider moving this bullet higher in the list (after "Configures which enrichments."  It looks like a mistake floating about on its own.

docs/index.asciidoc
```diff
@@ -415,6 +459,62 @@ deprecated[6.4.0, Replaced by <<plugins-{type}s-{plugin}-ssl_supported_protocols
...
+
+
+As an example, following configuration disables _all_ enrichments:
```
karenzone, line 493:
```suggestion
**Example:**
This configuration disables _all_ enrichments:
```

docs/index.asciidoc
mashhurs: TIL: I will do it.

docs/index.asciidoc
mashhurs: I think this simplification makes more sense.

```suggestion
|When ECS is enabled, even if `[event][original]` field does not already exist on the event being processed, this plugin's *default codec* ensures that the field is populated using the bytes as-processed.
```

mashhurs: Thanks @karenzone for the review. I have reviewed and applied your feedbacks/suggestions, simplified this sentence for your final review.

docs/index.asciidoc
```diff
@@ -239,6 +279,63 @@ Close Idle clients after X seconds of inactivity.
...
+| all               | _alias_ to include _all_ available enrichments (including additional
+                      enrichments introduced in future versions of this plugin)
+| none              | _alias_ to _exclude_ all available enrichments. Note that, _explicitly_
+                      defining <<codec>> with this option will not disable the `ecs_compatibility`,
+                      instead it relies on pipeline or codec `ecs_compatibility` configuration.
+|=======================================================================
+
+
+**Example:**
+This configuration disables _all_ enrichments:
```
karenzone, line 314:
```suggestion
**Example:**

This configuration disables _all_ enrichments:
```

karenzone (approved):
LGTM! Thanks for all the good work on this. 🎉

PR #466, opened 2023-01-13
Updates Netty to 4.1.87

andsel:
## Release notes
Update Netty to `4.1.87`

## What does this PR do?

With Netty's [PR 11732](https://github.com/netty/netty/pull/11732), Netty projects stops to re-package jars in `netty-all` dependency, so all the Netty deps used by the project has to be explicited listed, and this started with version `4.1.69`.
With version [`4.1.78`](https://netty.io/news/2022/06/14/4-1-78-Final.html) was shipped the [PR 12438](https://github.com/netty/netty/pull/12438/files) which in `SslContext` needs the unix socket domains, packaged in `netty-transport-native-unix-common`.

Keeps the dependencies updated.


mashhurs (approved):
LGTM!

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 6.4.4
```
mashhurs, line 1: This will be conflict with my recent [PR](https://github.com/logstash-plugins/logstash-input-beats/pull/464) but I will sort it out if you merge first.

build.gradle
```diff
@@ -15,6 +15,7 @@ targetCompatibility = 1.8
 
 String jacksonVersion = '2.13.3'
```
mashhurs, line 16: Not with this PR but we also need to update `jackson` dependencies to `2.14.1`.

PR #467, opened 2023-01-13
Fix doc formatting regression

karenzone:
Corrected formatting to resolve this formatting regression: 
Screen Shot 2023-01-13 at 3 30 37 PM


karenzone (commented):
Verified fix in local build

Issue #399, opened 2020-10-29
When `ssl_certificate_authorities` provided, `ssl_verify_mode => none` is ignored.


yaauie:
Follow-up to #300, which was closed and defined this as desired behaviour:

> The act of setting `certificate_authorities` should be an expression by the user/operator that they _want_ to only accept trusted clients. Maybe we need to improve the wording/docs here?
> 
> Alternately, why would you set `ssl_verify => none` _and_ also set _[`ssl_certificate_authorities`]_ ? Why would you say "Trust the things in _[`ssl_certificate_authorities`]_" but then tell the plugin to ignore trust relationships? What's the use case and intent?
>
> -- https://github.com/logstash-plugins/logstash-input-beats/issues/300#issuecomment-365687897

We need to make this experience better for our users.

### Option 1:

 - Reject `ssl_verify => none` when `ssl_certificate_authorities` is provided.

### Option 2:

 - WARN when encountering `ssl_verify => none` when `ssl_certificate_authorities` is provided.
 - Document that the setting is ignored in some cases (and define which specific cases).

jsvd:
> Alternately, why would you set ssl_verify => none and also set [ssl_certificate_authorities] ? Why would you say "Trust the things in [ssl_certificate_authorities]" but then tell the plugin to ignore trust relationships? What's the use case and intent?

I understand the argument, however the issue is that we default the verify mode to "force_peer" in the [SslContextBuiilder](https://github.com/logstash-plugins/logstash-input-beats/blob/master/src/main/java/org/logstash/netty/SslContextBuilder.java#L38). This means that `ssl_verify => none` with ssl_certificates_authorities will be stricter than setting `ssl_verify => peer`.

So I'm +1 on adding a warning for now and then going for Option 1 in a next major.

PR #470, opened 2023-03-28
Standardize SSL settings

edmocosta:
### What this PR does?
**Added the following SSL settings**
`ssl_enabled`: Enable/disable SSL
`ssl_client_authentication`: Controls the server’s behavior in regard to requesting a certificate from client connections
   
**Reviewed and deprecated the following SSL settings to comply with Logstash's [naming convention](https://github.com/elastic/logstash/issues/14905)**
`ssl` in favor of `ssl_enabled`
`ssl_verify_mode` in favor of `ssl_client_authentication`

**Other changes**

- Using `ssl_verify_mode` should keep the current behavior as it is. Once upgraded to `ssl_client_authentication`, it will validate and reject configurations with `ssl_certificate_authorities` set and `ssl_client_authentication = > none`.  The current `ssl_verify_mode` behavior is to silently ignore the `none` value and use `force_peer/required`.

- Fixed `ssl_peer_metadata` when `ssl_enabled => false`: https://github.com/logstash-plugins/logstash-input-beats/pull/470#discussion_r1153374813

- Added tests for the new and existing `SslContextBuilder` methods.

The behavior standardization across plugins, such as the accepted certificate formats, default values, etc will be tackled in future PRs.

---
Closes https://github.com/elastic/logstash/issues/14925
Closes https://github.com/logstash-plugins/logstash-input-beats/issues/399


lib/logstash/inputs/beats.rb
```diff
@@ -275,20 +259,39 @@ def target_codec_on_field?
...
+    certificate_authorities_configured?
+  end
+
+  def certificate_authorities_configured?
     @ssl_certificate_authorities && @ssl_certificate_authorities.size > 0
   end
 
   def client_authentication_metadata?
-    @ssl_peer_metadata && ssl_configured? && client_authentification?
+    @ssl_enabled && @ssl_peer_metadata && ssl_configured? && client_authentication_enabled?
```
edmocosta, line 278: I have added the `@ssl_enabled` here to fix an existing issue. When `ssl => false`, ` ssl_peer_metadata => true`, and `ssl_certificate_authorities` is configured, it tries to [extract the metadata ](https://github.com/logstash-plugins/logstash-input-beats/blob/main/lib/logstash/inputs/beats/message_listener.rb#L135) from the netty context and crashes as SSL is actually disabled.

src/main/java/org/logstash/netty/SslContextBuilder.java
```diff
@@ -29,25 +28,41 @@
...
+                case OPTIONAL:
+                    return ClientAuth.OPTIONAL;
+                case REQUIRED:
+                    return ClientAuth.REQUIRE;
+                default:
+                    throw new IllegalArgumentException(
+                            String.format("Unsupported SSL client authentication mode `%s`", this.name().toLowerCase())
+                    );
+            }
+        }
```
andsel, line 48: Maybe we could have a more explicit mapping without using the switch with:
```java
  NONE(ClientAuth.NONE),
  OPTIONAL(ClientAuth.OPTIONAL),
  REQUIRED(ClientAuth.REQUIRE);
  
  private final ClientAuth clientAuth;

  private SslClientVerifyMode(ClientAuth clientAuth) {
      this.clientAuth = clientAuth;
  }
   
  public ClientAuth toClientAuth() {
      return clientAuth;
  }
```

src/main/java/org/logstash/netty/SslContextBuilder.java
```diff
@@ -129,58 +162,34 @@ public static String[] getDefaultCiphers(){
...
-    }
 
     public SslContext buildContext() throws Exception {
-        io.netty.handler.ssl.SslContextBuilder builder = io.netty.handler.ssl.SslContextBuilder.forServer(sslCertificateFile, sslKeyFile, passPhrase);
-
         if (logger.isDebugEnabled()) {
-            logger.debug("Available ciphers: " + SUPPORTED_CIPHERS);
-            logger.debug("Ciphers:  " + Arrays.toString(ciphers));
+            logger.debug(String.format("Available ciphers: %s", SUPPORTED_CIPHERS));
+            logger.debug(String.format("Ciphers: %s", Arrays.toString(ciphers)));
```
andsel, line 177: We can leverage the log4j templating and avoid the move to String format


```suggestion
            logger.debug("Available ciphers: {}", SUPPORTED_CIPHERS);
            logger.debug("Ciphers: {}", Arrays.toString(ciphers));
```

src/main/java/org/logstash/netty/SslContextBuilder.java
```diff
@@ -129,58 +162,34 @@ public static String[] getDefaultCiphers(){
...
-            builder.trustManager(loadCertificateCollection(certificateAuthorities));
-            if(verifyMode == SslClientVerifyMode.FORCE_PEER) {
-                // Explicitly require a client certificate
-                builder.clientAuth(ClientAuth.REQUIRE);
-            } else if(verifyMode == SslClientVerifyMode.VERIFY_PEER) {
-                // If the client supply a client certificate we will verify it.
-                builder.clientAuth(ClientAuth.OPTIONAL);
+        if (isClientAuthenticationEnabled(verifyMode)) {
+            if (logger.isDebugEnabled()) {
+                logger.debug(String.format("Certificate Authorities: %s", Arrays.toString(certificateAuthorities)));
```
andsel, line 187:
```suggestion
                logger.debug("Certificate Authorities: {}", Arrays.toString(certificateAuthorities));
```

src/test/java/org/logstash/netty/SslContextBuilderTest.java
```diff
@@ -1,23 +1,76 @@
...
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.handler.ssl.ClientAuth;
+import io.netty.handler.ssl.SslContext;
 import org.hamcrest.core.Every;
+import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
+import static org.hamcrest.Matchers.*;
```
andsel, line 12: Avoid `*` import switch to explicit list

andsel: same for the following import of such kind

src/test/java/org/logstash/netty/SslContextBuilderTest.java
```diff
@@ -37,4 +89,141 @@ public void getDefaultCiphers() throws Exception
...
+        assertThat(sslContextBuilder.getVerifyMode(), is(SslClientVerifyMode.NONE));
+    }
+
+    private void assertSetClientAuthenticationThrowWithNoCerts(SslClientVerifyMode mode) {
+        final SslContextBuilder sslContextBuilder = createSslContextBuilder();
+        final String expectedMessage = "Certificate authorities are required to enable client authentication";
+
+        try {
+            sslContextBuilder.setClientAuthentication(mode, new String[0]);
+            Assert.fail("No exception was thrown");
```
andsel, line 139: All the other `Assert` methods are statically imported, so maybe we could use the same pattern and move to `fail()`

andsel (changes requested):
Overall seems good, but left some comments for improvement.

lib/logstash/inputs/beats.rb
```diff
@@ -275,20 +259,39 @@ def target_codec_on_field?
...
   end
 
   def client_authentication_required?
-    @ssl_verify_mode == "force_peer" 
+    @ssl_client_authentication && @ssl_client_authentication.downcase == "required"
```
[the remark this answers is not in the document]
edmocosta: Hi @andsel, thanks for reviewing! I don't see much gain as well so I will keep it as it's.

andsel (approved):
LGTM

PR #476, opened 2023-07-25
Document how to size the direct memory under expected event size

andsel:


docs/index.asciidoc
```diff
@@ -100,6 +100,31 @@ Setting direct memory too low decreases the performance of ingestion.
 
 NOTE: Be sure that heap and direct memory combined does not exceed the total memory available on the server to avoid an OutOfDirectMemoryError
 
+[id="plugins-{type}s-{plugin}-memory-sizing"]
+===== How to size the direct memory used
+
+To correctly size the direct memory to sustain the flow of incoming Beats connections, the medium size of the transmitted
+log lines and the batch size used by Beats (default to 2048), has to be known. For each connected client, a batch of events
+is read and due to the way the decompressing and decoding part works, it keeps two copies of the batch in memory.
```
andsel, line 108: Note for reviewer:
Beats decoding part keeps 2 copies of the buffer it's processing in memory:
- one for the decompression: https://github.com/logstash-plugins/logstash-input-beats/blob/339913cbf0268da82ab517cfdbf5e3c3ece99191/src/main/java/org/logstash/beats/BeatsParser.java#L232-L234
- the other is kep in the batch before lazily creating LogStash events:  https://github.com/logstash-plugins/logstash-input-beats/blob/339913cbf0268da82ab517cfdbf5e3c3ece99191/src/main/java/org/logstash/beats/V2Batch.java#L12

PR #479, opened 2023-09-06
acks: send sequence-0 ack for 0-window batches

yaauie:
Ensures that at least one ACK is sent in reply to each window, including those with 0-length, by emitting an empty batch when we receive a 0-window frame and sending a sequence-0 ACK in reply when we process an empty batch.

This enables upstream users of the protocol to (ab)use a 0-length window frame and an _ensured_ in-protocol response to validate that we are alive.

Note: 0-sequence ACKs are already used as application-level keep-alives between when the `ConnectionHandler` receives bytes and when `BeatsHandler` finishes processing a batch, and therefore should already be handled gracefully by any existing clients.


.ci/run.sh
```diff
@@ -5,6 +5,7 @@ env
 
 set -ex
 
+bundle exec rake test:java
```
yaauie, line 8: review note: java unit tests were not being triggered as part of the build 🤦🏼

jsvd (approved):
LGTM, I tested locally with a modified lumberjack client. I could observe that with this patch the client immediately received an ack after sending an zero window frame, instead of waiting a long time for the ack from the idlestatehandler.

The script used, that takes code from the original lumberjack ruby client, can be found here: https://gist.github.com/jsvd/b86ea9d2f16a868c1460d4e12145a35f#file-beats_empty_writer-rb

Issue #480, opened 2023-09-25
Remove logstash-forwarder from integration testing


andsel:
[Logstash-forwarder ](https://github.com/elastic/logstash-forwarder) has been completely supplanted by Beats and ElasticAgent, so there is no need to the keep testing on it like in https://github.com/logstash-plugins/logstash-input-beats/blob/339913cbf0268da82ab517cfdbf5e3c3ece99191/spec/integration/logstash_forwarder_spec.rb .

- [x] remove the testing code being sure that there isn't any loose in code coverage terms, that all meaningful source paths for Beats keeps tested.
- [x] update the fixture preparation test script to don't download it: https://github.com/logstash-plugins/logstash-input-beats/blob/339913cbf0268da82ab517cfdbf5e3c3ece99191/lib/tasks/test.rake#L34-L43

PR #481, opened 2023-09-25
Remove LSF from integration testing

roaksoax:
Closes: #480


lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -181,21 +181,13 @@ def set_nested(hash, field_name, value)
 
     private
     def extract_target_field(hash)
-      if from_filebeat?(hash)
-        hash.delete(FILEBEAT_LOG_LINE_FIELD).to_s
-      elsif from_logstash_forwarder?(hash)
-        hash.delete(LSF_LOG_LINE_FIELD).to_s
-      end
+      hash.delete(FILEBEAT_LOG_LINE_FIELD).to_s
```
andsel, line 184: The tests fails because expects a `nil` instead received an empty string.
I think the code to satisfy again the CI is
```suggestion
      return nil if !from_filebeat?(hash)
      hash.delete(FILEBEAT_LOG_LINE_FIELD).to_s  
```

andsel (dismissed):
Left a comment to make the CI tests green again

andsel (approved):
LGTM

yaauie:
⚠️ This patch-level "fix" removes support for decoding events received from the Lumberjack input.

PR #485, opened 2023-11-07
Switch from push mode to pull

andsel:
## Release notes

To control spikes of memory usage, the incoming message reads switched from push mode to pull, so that the plugin has control on rate of ingestion and it's not the determined by the clients.

## What does this PR do?

Changes the way the plugin pulls data and handle incoming connections to avoid many clients pushing more data than what it can handle.
Switches to pull read instead of push read. Instead to respond to every buffer that a connection present, it's the plugin that actively grab the byte buffer to process.

This work is based on the existing #410 #475.

Provides a counter measure to avoid that clients decide on the memory of the local process, pushing more and more data also if the process can't cope with the flow rate of allocations.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files (and/or docker env variables)
- [x] I have added tests that prove my fix is effective or that my feature works

## Author's Checklist

- [x] run with some real beats instances

### DIFF
--- a/src/main/java/org/logstash/beats/Server.java
+++ b/src/main/java/org/logstash/beats/Server.java
@@ -134,6 +134,7 @@ public class Server {
                              new IdleStateHandler(localClientInactivityTimeoutSeconds, IDLESTATE_WRITER_IDLE_TIME_SECONDS, localClientInactivityTimeoutSeconds));
             pipeline.addLast(BEATS_ACKER, new AckEncoder());
             pipeline.addLast(CONNECTION_HANDLER, new ConnectionHandler());
+            pipeline.addLast(new FlowLimiterHandler());
             pipeline.addLast(beatsHandlerExecutorGroup, new BeatsParser(), new BeatsHandler(localMessageListener));
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/logstash/beats/Server.java
```diff
@@ -134,6 +134,7 @@ public void initChannel(SocketChannel socket){
                              new IdleStateHandler(localClientInactivityTimeoutSeconds, IDLESTATE_WRITER_IDLE_TIME_SECONDS, localClientInactivityTimeoutSeconds));
             pipeline.addLast(BEATS_ACKER, new AckEncoder());
             pipeline.addLast(CONNECTION_HANDLER, new ConnectionHandler());
+            pipeline.addLast(new FlowLimiterHandler());
```
jsvd, line 137: Do we have a reason why this is placed in this location of the pipeline? would it make sense to be first instead? Or right after the ssl handler?

andsel: Moved in the same position it was from the originating PR #475, but your point is correct. It should be the first handler in the Netty pipeline

PR #486, opened 2023-11-20
Revert "Remove LSF from integration testing and core references (#481)"

roaksoax:
### This reverts commit 045626b8af8cec6ca85d1e9a577ff05ace1ab231.
Restores accidentally removed source pieces. The case where events from Lumberjack come with `line` field is broken now and this PR restores the feature. We will address removing LSF integration tests with separate PR.


mashhurs (approved):
LGTM!

PR #488, opened 2023-12-21
Remove LSR test cases, add Lumberjack test cases and format the classes.

### LINKED ISSUE #487: Add test cases for Lumberjack events and remove LSF tests
### Description
We recently [reverted LSF integration test removal commit](https://github.com/logstash-plugins/logstash-input-beats/pull/486) due to a reason, Lumberjack events weren't parsed correctly with JSON codec. With this task, we need to remove LSF integration tests and add test cases for Lumberjack event to make the consistency.

mashhurs:
### Description
Logstash forwarder is replaced by beats and elastic agent, we don't need to keep the tests against it. At the same time, we need to keep the Lumberjack compatibility as some users are still using Lumberjack output to Beats input architecture.

This change:
- removes LSR test cases
- adds/adjusts Lumberjack unit test case
- formats Java sources

- Closes #487


spec/inputs/beats/message_listener_spec.rb
```diff
@@ -199,7 +199,7 @@ def flush(&block)
       end
     end
 
-    context "when the message is from LSF" do
+    context "when the message is from Lumberjack" do
       let(:message) { MockMessage.new("abc", { "line" => "hello world", '@metadata' => {} } )}
```
mashhurs, line 203: review note: we keep unit test against Lumberjack and if we need integration tests we better place in [the core](https://github.com/elastic/logstash/blob/main/qa/integration/specs/beats_input_spec.rb) but since we are not actively maintaining Lumberjack and pushing LS-LS comm with new integration plugin, I don't think it is mandatory now.

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -7,23 +7,28 @@
 
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.util.Objects;
```
mashhurs, line 10: review note: Java source changes are standardization and formats. No logic changes.

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -7,23 +7,28 @@
...
 import javax.net.ssl.SSLHandshakeException;
 
 public class BeatsHandler extends SimpleChannelInboundHandler<Batch> {
-    private final static Logger logger = LogManager.getLogger(BeatsHandler.class);
+
+    private final static Logger LOGGER = LogManager.getLogger(BeatsHandler.class);
```
roaksoax, line 15: why the change since they are not global variables? Seems that this are not relevant to the actual changes needed?

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -139,21 +142,13 @@ private String format(String message) {
...
-
-        String remotehost;
-        if(remote != null) {
-            remotehost = remote.getAddress().getHostAddress() + ":" + remote.getPort();
-        } else{
-            remotehost = "undefined";
-        }
-
-        return "[local: " + localhost + ", remote: " + remotehost + "] " + message;
+        final String localhost = Objects.isNull(local)
```
roaksoax, line 145: again, seem unrelated changes?

spec/integration/logstash_forwarder_spec.rb
```diff
@@ -1,108 +0,0 @@
-# encoding: utf-8
```
roaksoax, line 1: shouldn't be build a version of tests for lumberjack?

roaksoax (changes requested):

src/main/java/org/logstash/beats/BeatsHandler.java
mashhurs: I was touching while formatting the file but since they non-related reverted back to origin state, this file change no longer included.

spec/integration/logstash_forwarder_spec.rb
mashhurs: I would support adding plugin level changes here and cover Lumberjack integration tests in the core like [we have beats](https://github.com/mashhurs/logstash/blob/19637143e6ef24b7b4009e71b1fa32b04b92c942/qa/integration/specs/beats_input_spec.rb#L26). That would cover `lumberjack -> beats` and `lumberjack -> lumberjack` use cases. I don't see any lumberjack integration tests but we [suggest it to LS-to-LS comm](https://www.elastic.co/guide/en/logstash/current/ls-to-ls-lumberjack.html) 🤦

roaksoax (approved):
lgtm!

PR #490, opened 2024-01-06
Introduces expert only event loop thread count setting.

mashhurs:
### Description
By default, `netty` event loop creates `available_processors * 2` thread count. This may sometimes lead a resource problem, especially when multiple beats-input pipelines are defined in a single machine.
This PR introduces a configuration to limit a number of threads for netty event loop threads.

### How to locally test
- unfortunately I couldn't find a better way to add a unit test for this feature
- pull the change and run pipeline with following config
- connect, reconnect multiple times the client to the downstream beats-input.
- check the threads (with thread dump tools or ls diagnostics tool) that two (or defined) threads will created for netty event loop named as `nioEventLoopGroup-M-N`

```
input {
    beats {
        codec => json
        port => 5001
        event_loop_threads => 2
        executor_threads => 2
    }
}
output {
    stdout { codec => rubydebug }
}
```

image


docs/index.asciidoc
```diff
@@ -339,6 +340,18 @@ input {
 }
 --------------------------------------------------
 
+[id="plugins-{type}s-{plugin}-event_loop_threads"]
+===== `event_loop_threads`
+
+* Value type is <<number,number>>
+* Defaults to 0.
```
robbavey, line 347: When setting `0`, the actual default is `available_processors * 2`

docs/index.asciidoc
```diff
@@ -339,6 +340,18 @@ input {
...
+
+Most of the cases do not require touching this configuration.
```
robbavey, line 349:
```suggestion
This is an expert-level setting, and generally should not need to be set
```

robbavey (changes requested):
Code LGTM, some doc changes requested

docs/index.asciidoc
mashhurs, line 347:
```suggestion
* Defaults to 0.

When setting `0`, the actual default is `available_processors * 2`
```

mashhurs: Updated.

robbavey (approved):
LGTM

PR #500, opened 2024-08-08
Fix the abnormal case where codec is not available to accept the event.

### LINKED ISSUE #499: Crash with `(NoMethodError) undefined method 'accept' for nil:NilClass`.
**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`)
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
4. How was the Logstash Plugin installed

**JVM** (e.g. `java -version`):

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`)
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc).
3. Value of the `JAVA_HOME` environment variable if set.

**OS version** (`uname -a` if on a Unix-like system):

**Description of the problem including expected versus actual behavior**:

**Steps to reproduce**:

Please include a *minimal* but *complete* recreation of the problem,
including (e.g.) pipeline definition(s), settings, locale, etc.  The easier
you make for us to reproduce it, the more likely that somebody will take the
time to look at it.

 1.
 2.
 3.

**Provide logs (if relevant)**:
```
{"level":"INFO","loggerName":"org.logstash.beats.BeatsHandler","timeMillis":1721298048946,"thread":"defaultEventExecutorGroup-78-6","logEvent":{"message":"[local: 0.0.0.0:3511, remote: undefined] Handling exception: org.jruby.exceptions.NoMethodError: (NoMethodError) undefined method `accept' for nil:NilClass (caused by: org.jruby.exceptions.NoMethodError: (NoMethodError) undefined method `accept' for nil:NilClass)"}}
{"level":"WARN","loggerName":"io.netty.channel.AbstractChannelHandlerContext","timeMillis":1721298048946,"thread":"defaultEventExecutorGroup-78-6","logEvent":{"message":"Failed to submit an exceptionCaught() event."}}
{"level":"WARN","loggerName":"io.netty.channel.AbstractChannelHandlerContext","timeMillis":1721298048946,"thread":"defaultEventExecutorGroup-78-6","logEvent":{"message":"The exceptionCaught() event that was failed to submit was:"}}
```

mashhurs:
### Description

Fixes the `undefined method 'accept'` error. Full error log:
```
{"level":"INFO","loggerName":"org.logstash.beats.BeatsHandler","timeMillis":1721298048946,"thread":"defaultEventExecutorGroup-78-6","logEvent":{"message":"[local: 0.0.0.0:3511, remote: undefined] Handling exception: org.jruby.exceptions.NoMethodError: (NoMethodError) undefined method `accept' for nil:NilClass (caused by: org.jruby.exceptions.NoMethodError: (NoMethodError) undefined method `accept' for nil:NilClass)"}}
{"level":"WARN","loggerName":"io.netty.channel.AbstractChannelHandlerContext","timeMillis":1721298048946,"thread":"defaultEventExecutorGroup-78-6","logEvent":{"message":"Failed to submit an exceptionCaught() event."}}
{"level":"WARN","loggerName":"io.netty.channel.AbstractChannelHandlerContext","timeMillis":1721298048946,"thread":"defaultEventExecutorGroup-78-6","logEvent":{"message":"The exceptionCaught() event that was failed to submit was:"}}
```

- Closes #499


lib/logstash/inputs/beats/message_listener.rb
```diff
@@ -49,11 +49,18 @@ def onNewMessage(ctx, message)
         @nocodec_transformer.transform(event)
         @queue << event
       else
-        codec(ctx).accept(CodecCallbackListener.new(target_field,
-                                                    hash,
-                                                    message.getIdentityStream(),
-                                                    @codec_transformer,
-                                                    @queue))
+        current_codec = codec(ctx)
+        if current_codec
```
mashhurs, line 53: review note: this condition allows  `BeatsHandler#channelRead0()` to continue without crashing on `messageListener.onNewMessage(ctx, message);`. But since reproducing the issue is very hard (I couldn't with multiple tries), I have no idea what will happen next. So focusing on only solving the `(NoMethodError) undefined method 'accept' for nil:NilClass` error.

My concern is, since connection is not available, [ack operation (`BeatsHandler#writeAck`)](https://github.com/logstash-plugins/logstash-input-beats/blob/main/src/main/java/org/logstash/beats/BeatsHandler.java#L131) or `ctx.flush()` still would fail.

src/main/java/org/logstash/beats/Server.java
```diff
@@ -51,14 +54,15 @@ public Server listen() throws InterruptedException {
...
+        bossGroup = new NioEventLoopGroup(eventLoopThreadCount); // TODO: add a config to make it adjustable, no need many threads
         workGroup = new NioEventLoopGroup(eventLoopThreadCount);
         try {
             logger.info("Starting server on port: {}", this.port);
 
             beatsInitializer = new BeatsInitializer(messageListener, clientInactivityTimeoutSeconds, executorThreadCount);
 
             ServerBootstrap server = new ServerBootstrap();
-            server.group(workGroup)
+            server.group(bossGroup, workGroup)
```
mashhurs, line 65: Till now, we were using workgroup for handling incoming socket connections and there was no guarantee socket handlers got terminated first. With this separation we will have two major improvements:
- make sure we terminate handlers, which don't push to netty internal tasks queue
- make thread count configurable: no need many threads as it accepts and sends to work group

References:
- https://dzone.com/articles/building-a-performant-application-using-netty
- https://netty.io/wiki/user-guide-for-4.x.html#wiki-h3-16

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -7,22 +7,28 @@
...
 import javax.net.ssl.SSLHandshakeException;
 
 public class BeatsHandler extends SimpleChannelInboundHandler<Batch> {
     private final static Logger logger = LogManager.getLogger(BeatsHandler.class);
+    private final static String executorTerminatedMessage = "event executor terminated";
+
     private final IMessageListener messageListener;
     private ChannelHandlerContext context;
 
+    private final AtomicBoolean isQuitePeriod = new AtomicBoolean(false);
```
andsel, line 22:
```suggestion
    private final AtomicBoolean isQuietPeriod = new AtomicBoolean(false);
```

lib/logstash/inputs/beats/message_listener.rb
andsel: > By better exception I mean, BeatsHandler#exceptionCaught fails to send super.exceptionCaught(ctx, cause); but gets was submitted.... handler didn't handle last exception cyclic message.

Please could you describe better this sentence, I don't grasp it. In particular under which condition `exceptionCaught` fails to invoke `super.exceptionCaught`. From actual code https://github.com/logstash-plugins/logstash-input-beats/blob/5ff649a7561f6e25a67b2c373ca265cd1a81a05d/src/main/java/org/logstash/beats/BeatsHandler.java#L88-L101 it could fail if:
- `isNoisyExecption()` fails itself
or
- the exception is not nosy (no connection reset by the peer) _and_
- `extractCause` fails itself

andsel (commented):
As I can understand the idea of this PR is to use bossGroup and workerGroup, so that the boss group can be shutdown before the worker one.

It's not completely clear to me the root of the problem when we use the same event loop group to do both IO on the acceptor socket and IO on the client socket.
The null reference on the codec, happens because the `connections_list` doesn't have anymore the `ConnectionState` for a context (so essentially the Channel). That's originated from the call to `onConnectionClose` of message listener done by `BeatsHandler.channelInactive()`.
Given that the IO for a channel is managed always by the same EventLoop thread the null reference could be imagined in 2 situations:
- income flow (client to server direction): the closing of a channel was processed before processing other data frames from the same channel. As per the constraint specified above (always same thread process data for same channel, so it's sort of serialisation) this case shouldn't be realised.
- outcome flow: the server is writing an ACK to the client which is now disconnected. In this case would be enough that the BeatsHandler checks the channel is active before writing.

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -41,29 +47,22 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+        if (logger.isDebugEnabled()) {
             logger.debug(format("Received a new payload"));
         }
         try {
-            if (batch.isEmpty()) {
-                logger.debug("Sending 0-seq ACK for empty batch");
-                writeAck(ctx, batch.getProtocol(), 0);
-            }
-            for (Message message : batch) {
+            if (isQuitePeriod.get()) {
```
mashhurs, line 54:
```suggestion
            if (isQuietPeriod.get()) {
```

src/main/java/org/logstash/beats/BeatsHandler.java
```diff
@@ -93,19 +92,47 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
...
                     logger.info(format("Handling exception: " + cause + " (caused by: " + realCause + ")"), cause);
                 } else {
                     logger.info(format("Handling exception: " + cause + " (caused by: " + realCause + ")"));
                 }
-                super.exceptionCaught(ctx, cause);
+                // when execution tasks rejected, no need to forward the exception to netty channel handlers
+                if (cause instanceof RejectedExecutionException) {
+                    // we no longer have event executors available since they are terminated, mostly by shutdown process
+                    if (Objects.nonNull(cause.getMessage()) && cause.getMessage().contains(executorTerminatedMessage)) {
+                        this.isQuitePeriod.compareAndSet(false, true);
```
mashhurs, line 104:
```suggestion
                        this.isQuietPeriod.compareAndSet(false, true);
```

lib/logstash/inputs/beats/message_listener.rb
mashhurs: It means any exception isn't handled by overridden `BeatsHandler#exceptionCaught` method causes internal netty exception and we see exception flood (exception content has `was submitted.... but didn't handle last exception`), usually raised in channels. `BeatsHandler#exceptionCaught` needs to either let netty know the exception by calling `super.exceptionCaught(ctx, cause);` or swallow it if expected behavior (`isNoisyException` or `RejectedExecutionException` for this case).

mashhurs:
> As I can understand the idea of this PR is to use bossGroup and workerGroup, so that the boss group can be shutdown before the worker one.
> 
> It's not completely clear to me the root of the problem when we use the same event loop group to do both IO on the acceptor socket and IO on the client socket. 

Although shutdown requested, incoming requests keep coming and netty places requests into its task queue. netty terminates its worker group and boss group threads but no guarantee that boss group terminates first unless we intentionally force to terminate it first. So, `BeatHandler#exceptionCaught` receives `RejectedExecutionException` because no executor thread left to continue the task. `BeatHandler#channelRead0` keeps receiving incoming messages and causes 
- `undefined method 'accept'`
- internal netty `failed to submit exceptionCaught` exception flood

```
// example logs
{"level":"WARN","loggerName":"io.netty.channel.AbstractChannelHandlerContext","timeMillis":1721298048741,"thread":"defaultEventExecutorGroup-78-1","logEvent":{"message":"Failed to submit an exceptionCaught() event."}}
{"level":"WARN","loggerName":"io.netty.channel.AbstractChannelHandlerContext","timeMillis":1721298048741,"thread":"defaultEventExecutorGroup-78-1","logEvent":{"message":"The exceptionCaught() event that was failed to submit was:"}}
```

>The null reference on the codec, happens because the `connections_list` doesn't have anymore the `ConnectionState` for a context (so essentially the Channel). That's originated from the call to `onConnectionClose` of message listener done by `BeatsHandler.channelInactive()`. Given that the IO for a channel is managed always by the same EventLoop thread the null reference could be imagined in 2 situations:
> 
> * income flow (client to server direction): the closing of a channel was processed before processing other data frames from the same channel. As per the constraint specified above (always same thread process data for same channel, so it's sort of serialisation) this case shouldn't be realised.
> * outcome flow: the server is writing an ACK to the client which is now disconnected. In this case would be enough that the BeatsHandler checks the channel is active before writing.

andsel:
In https://github.com/logstash-plugins/logstash-input-beats/blob/5ff649a7561f6e25a67b2c373ca265cd1a81a05d/src/main/java/org/logstash/beats/Server.java#L61 we use the `group` method as defined in Javadoc https://netty.io/4.1/api/io/netty/bootstrap/ServerBootstrap.html#group-io.netty.channel.EventLoopGroup- it uses the same event loop group both for acceptor and client sockets, so in this case there isn't any distinction between boss group and worker group because are the same,

> netty terminates its worker group and boss group threads but no guarantee that boss group terminates first unless we intentionally force to terminate it first

how it's possible that `BeatHandler.channelRead0` continue receiving tasks if the only event loop group is in shutdown?

src/main/java/org/logstash/beats/Server.java
```diff
@@ -83,13 +87,30 @@ public void stop() {
...
+                workGroup.shutdownGracefully();
             }
             if (beatsInitializer != null) {
                 beatsInitializer.shutdownEventExecutor();
             }
+            // we wait for workGroup shutdown this time
+            if (workGroup != null && !workGroup.isShutdown()) {
+                // calling again shutdownGracefully doesn't hurt that it returns terminationFuture if shutdown is under process
+                workGroup.shutdownGracefully().sync();
+            }
```
andsel, line 113: Not clear why after  the previous call `workGroup.shutdownGracefully` we need a second one.

`shutdownGracefully` at first thing sets its status in `ST_SHUTTING_DOWN`.  Maybe it's due to the default quiet period (2 seconds by default), so that is could happen that after the first call to `workGroup.shutdownGracefully` another task fall in the task queue before the quiet period elapses, and the shutdown status is reset.  If this is the case, we could set 0 as quiet period or `sync` on the first call.

andsel (commented):
Left a comment on the shutdown sequence of the event loop groups.

src/main/java/org/logstash/beats/Server.java
mashhurs: Yup, thanks for catching up, forgot to push my last local update.
Worker shutdown doesn't need to necessarily be in this order, simply boss -> group -> event executors termination is a technically correct way. I have pushed the change.

src/main/java/org/logstash/beats/Server.java
```diff
@@ -160,6 +177,18 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
         public void shutdownEventExecutor() {
             try {
                 idleExecutorGroup.shutdownGracefully().sync();
+
+                // DefaultEventExecutorGroup internally executes numbers of SingletonEventExecutor
```
andsel, line 181:
```suggestion
                // DefaultEventExecutorGroup internally executes numbers of SingleThreadEventExecutor
```

src/main/java/org/logstash/beats/Server.java
```diff
@@ -160,6 +177,18 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
...
+                // try to gracefully shut down every thread if they have unacked pending batches (pending tasks)
+                for (final EventExecutor eventExecutor : beatsHandlerExecutorGroup) {
+                    if (eventExecutor instanceof SingleThreadEventExecutor) {
+                        final SingleThreadEventExecutor singleExecutor = (SingleThreadEventExecutor) eventExecutor;
+                        if (singleExecutor.pendingTasks() > 0) {
+                            singleExecutor.shutdownGracefully().sync();
+                        }
+                    }
+                }
```
andsel, line 190: Why we need this if the `DefaultEventExecutorGroup.shutdownGracefully` already does it, but without considering the `pendingTasks`?
As I can understand we split the shutdown of the `beatsHandlerExecutorGroup` in two phases:
- in the first we shutdown the executors that has pending tasks
- in the second the executors remained

Why we need it?
If we need it, could be better to extract this loop in a method that is more explicit of the intention, like ofr example `shutdownExecutorsWithPendingTasks` ?

https://github.com/netty/netty/blob/netty-4.1.109.Final/common/src/main/java/io/netty/util/concurrent/MultithreadEventExecutorGroup.java#L161-L166

andsel (changes requested):
I've left a request why we need to shutdown the `beatsHandlerExecutorGroup` in such peculiar way and eventually left a suggestion to make the intention more explicit.

src/main/java/org/logstash/beats/Server.java
mashhurs: - Separated into `shutdownEventExecutorsWithPendingTasks` method. 
- `beatsHandlerExecutorGroup.shutdownGracefully().sync();` is a concurrent approach (we don't know which thread occupies CPU and runs). Asking each busy/pending executor to shutdown gives opportunity to process the events it has (as much as prevents data loss), especially if quiet time rollout (see https://netty.io/4.1/api/io/netty/util/concurrent/EventExecutorGroup.html#shutdownGracefully-long-long-java.util.concurrent.TimeUnit-).

andsel (approved):
LGTM

PR #509, opened 2024-12-03
Remove jar-dependencies dependency

robbavey:


donoghuc (approved):
I believe the `7.current` cell is expected to be failing due to where we are at in release cycle as `7.17.27-SNAPSHOT` is not yet available. Other tests are 🍏

Issue #519, opened 2025-09-04
Fix hanging tests


donoghuc:
Tests are hanging/retrying for 1.5hrs resulting in being terminated. This is wasting limited travis resources. Track down the source of the bug and get tests back to a state where they wont hang until killed. 

**Failure history**:
 
- https://app.travis-ci.com/github/logstash-plugins/logstash-input-beats/builds

PR #520, opened 2025-09-05
Use 8.19 series for filebeat integration tests

donoghuc:
This commit updates the filebeat integration tests to use a newer version. Note that for ci we use x86_64 for linux and for local testing arm on mac.

Implements https://github.com/logstash-plugins/logstash-input-beats/issues/519


robbavey (approved):
LGTM

PR #521, opened 2025-09-05
Use 8.19 series for filebeat integration tests

donoghuc:
This commit updates the filebeat integration tests to use a newer version. Note that for ci we use x86_64 for linux and for local testing arm on mac.


donoghuc:
Tests are green!

PR #523, opened 2025-10-13
[docs] Avoid unnecessary updates to the docs

colleenmcginnis:
Avoid unnecessary updates to the generated HTML used to generate [these docs](https://www.elastic.co/docs/reference/logstash/plugins/plugins-inputs-beats). This approach allows us to use a built-in AsciiDoc attribute for the year so the example doesn't become terribly old, but still limits the noise in the diff when updating the docs.

cc @karenzone


karenzone:
Published at [v7.0.4](https://rubygems.org/gems/logstash-input-beats/versions/7.0.4-java)

PR #527, opened 2026-01-26
Repacking

mashhurs:
TODOs:
- If approach looks good, I will version bump and add changelogs
- After landing this change, we need to think about backporting to 8.x stack eligible branch.


build.gradle
```diff
@@ -26,25 +26,24 @@ repositories {
...
     implementation "io.netty:netty-handler:${nettyVersion}"
-    implementation "io.netty:netty-transport-native-unix-common:${nettyVersion}" // this is needed from Netty 4.1.78.Final on
     implementation 'org.javassist:javassist:3.24.0-GA'
-    testImplementation "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
-    testImplementation "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
-    testImplementation "com.fasterxml.jackson.core:jackson-databind:${jacksonDatabindVersion}"
-    testImplementation "com.fasterxml.jackson.module:jackson-module-afterburner:${jacksonVersion}"
+
     compileOnly "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
     compileOnly "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
```
donoghuc: Wouldnt we only need jackson-core (the other jackson deps are transitive).

PR #537, opened 2026-04-16
Bind in register

### LINKED ISSUE #538: Bind to port during `#register`
When we fail to bind to the port, we should propagate the failure to prevent the pipeline from starting.

With this netty-based plugin, the pattern should be:
 - bind in `#register` with the server channel set to _disable_ auto-read
 - enable auto-read in `#run` once the plugin has a reference to the queue writer

Relates: elastic/logstash#18711

yaauie:
This Pr is best reviewed commit-by-commit. Each commit passes tests with `ELASTIC_STACK_VERSION=9.current`, which currently resolves to 9.3.3.

Moves the binding of the port into `#register` (with the listening channel's auto-read disabled), and then activates the auto-read in `#run` after the `MessageListener` with access to the queue is available.

Note: migrates from a blocking `Server#listen` to `Server#run(MessageListener)` in order to remain descriptive.

Meta: elastic/logstash#18711


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.0.9
+  - When configured to use a port that is already in use, the failure is now propagated to the pipeline
```
yaauie, line 2:
```suggestion
  - When configured to use a port that is already in use, the failure is now propagated to the pipeline [#537](https://github.com/logstash-plugins/logstash-input-beats/pull/537)
```

spec/support/logstash_test.rb
```diff
@@ -30,6 +30,37 @@ def certificate
     def random_port
       rand(2000..10000)
     end
+
+    ##
+    # Returns the IP address of an interfaace we own that is neither loopback nor multicast.
```
andrewvc, line 35: Type `s/interfaace/interface/`

lib/logstash/inputs/beats.rb
```diff
@@ -202,6 +202,12 @@ def register
     @logger.info("Starting input listener", :address => "#{@host}:#{@port}")
 
     @server = create_server
+    begin
+      @server.bind()
+      ObjectSpace.define_finalizer(self, self.class._server_shutdown_proc(@server))
```
andrewvc, line 207: Can you bring me up to speed on this pattern? I'm assuming that while we have `#run/#stop` there's something wrong with `#close`? We do try to call it in LS with [java_pipeline.rb](https://github.com/elastic/logstash/blob/main/logstash-core/lib/logstash/java_pipeline.rb#L258)

Even if so, it feels weird to me to link this with object lifecycle, feels like also releasing the port on `stop` is a good idea. 

AFAICT there are no guarantees a finalizer will ever run, see the [MRI Docs](https://docs.ruby-lang.org/en/master/ObjectSpace.html) . Relevant quote:

> Note that finalization can be unpredictable and is never guaranteed to be run except on exit.

Maybe JRuby has different behavior, but I couldn't find any sources for that. It seems unlikely given that `finalize` was [deprecated for removal in Java 18](https://openjdk.org/jeps/421)

IMO the best fix, probably too complex, would be to add `#unregister` to Logstash core and add support for this plugin. I think that's probably not worth the effort however.

A better middle-ground I think would be to:

1. Keep the bind in `#register`
2. Unbind in `#stop`
3. Set a timer in `#register` to unbind after say 1 minute if `#run` was not called
4. In the weird case where somehow there is a long delay between `#register` and `#run` allow `#run` to rebind

It's awkward but IMHO more robust and doesn't depend on shaky parts of the runtime.

lib/logstash/inputs/beats.rb
```diff
@@ -375,4 +379,21 @@ def resolve_enriches
 
     return @enrich
   end
+
+  # returns a proc to close the provided server without holding any other refs
+  # https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/
```
andrewvc, line 384: TIL!

src/main/java/org/logstash/beats/Server.java
```diff
@@ -64,25 +68,38 @@ public Server listen() throws InterruptedException {
...
-        return this;
+    public void run(final IMessageListener messageListener) throws InterruptedException {
+        synchronized (this) {
+            if (serverChannel == null) {
+                bind();
+            }
+        }
+        try {
+            beatsInitializer.setMessageListener(messageListener);
+            serverChannel.config().setAutoRead(true);
```
andrewvc, line 98: I had to lookup this function to determine what it did, worth a comment IMHO

```suggestion
            // Start accepting/reading connections now that we have a listener
            serverChannel.config().setAutoRead(true);
```

src/main/java/org/logstash/beats/Server.java
```diff
@@ -64,25 +68,38 @@ public Server listen() throws InterruptedException {
...
+            serverChannel.closeFuture().sync();
```
andrewvc, line 99:
```suggestion
            // Block until the server channel is closed.
            serverChannel.closeFuture().sync();
```

lib/logstash/inputs/beats.rb
andrewvc: As an update @yaauie and I spoke about this and agreed to move it to `#close`

andrewvc (dismissed):

spec/support/logstash_test.rb
yaauie:
```suggestion
    # Returns the IP address of an interface we own that is neither loopback nor multicast.
```

lib/logstash/inputs/beats.rb
```diff
@@ -202,6 +202,11 @@ def register
     @logger.info("Starting input listener", :address => "#{@host}:#{@port}")
 
     @server = create_server
+    begin
+      @server.bind()
+    rescue java.net.BindException => bind_exception
+      fail LogStash::ConfigurationError, "could not bind to #{@host}:#{@port}; #{bind_exception.message}"
```
donoghuc, line 209: Same question as https://github.com/logstash-plugins/logstash-input-http/pull/221#discussion_r3516075645

lib/logstash/inputs/beats.rb
```diff
@@ -211,9 +216,7 @@ def create_server
   end
 
   def run(output_queue)
-    message_listener = MessageListener.new(output_queue, self)
-    @server.setMessageListener(message_listener)
-    @server.listen
+    @server.run(MessageListener.new(output_queue, self))
   end # def run
 
   def stop
```
donoghuc, line 223: I see this stop method, but no close... Should there be a method like:
```
def close
  @server.stop unless @server.nil?
end
```
Which would match the http input plugin? https://github.com/logstash-plugins/logstash-input-http/pull/221/changes#diff-744e810bc24ab2f25364a478d9b956e541cb4b302670a538f0f49930311f5a0dR214

spec/support/logstash_test.rb
```diff
@@ -30,6 +30,37 @@ def certificate
...
+    # Yields block with a port that is unavailable
+    # @yieldparam port [Integer]
+    # @yieldreturn [Object]
+    # @return [Object]
+    def with_bound_port(host:"::", port:0, &block)
+      server = TCPServer.new(host, port)
+
+      return yield(server.local_address.ip_port)
+    ensure
+      server.close
```
donoghuc:
```suggestion
      server&.close
```
nit: another parity thing with http input plugin https://github.com/logstash-plugins/logstash-input-http/pull/221/changes#diff-9a02fee1a7f56b7b530d628aad25c5d3e0c9ccb6f2e5a70bf7e109eade70ce1dR29

donoghuc (changes requested):
Couple small parity things with http. Otherwise looks good!

lib/logstash/inputs/beats.rb
yaauie: I don't know why the http input implements `#close` or if anything even can call it outside of tests -- inputs hook into [`Inputs::Base#stop`](https://github.com/elastic/logstash/blob/v9.4.3/logstash-core/lib/logstash/inputs/base.rb#L90-L96) for stopping and the input base class doesn't _have_ a `#close` method.

lib/logstash/inputs/beats.rb
yaauie:
```suggestion
      @server.stop rescue nil
      fail LogStash::ConfigurationError, "could not bind to #{@host}:#{@port}; #{bind_exception.message}"
```

lib/logstash/inputs/beats.rb
donoghuc: ah, ok yeah maybe its a test setup thing. ty

donoghuc (approved):

PR #578, opened 2026-08-27
[Test] Added logs to state when download phase terminate and start un…

andsel:
Added logs to state when download phase terminate and start unpacking plus SHA512 verification of the downloaded artifact.


donoghuc (approved):
I think this will help us figure out what is going on here and how to guard against it. Maybe a retry of some sort. Lets see what this tells us.

